Issue #76, opened 2015-03-30
user switching is broken if 2 DMs are installed


NiceandGently:
In Code matesession checks first for MDM, than for GDM and at least for lightdm.
In case of GDM and lightdm is installed, and lightdm is active, mate-session doesn't check if lightdm is running and produce this error message.
```
** (gdmflexiserver:1616): WARNING **: Unable to create transient display: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.DisplayManager was not provided by any .service files
GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.DisplayManager was not provided by any .service filessys
```
Note, on systems with more than one user it is very often the case that gnome-shell and MATE is installed. It is imposssible to install gnome-shell without gdm, in result you have 2 DM's installed, because most default installations in distros has lightdm as default DM.
This is reproducible with Jessie too!

PR #77, opened 2015-03-30
use lightdm as second condition for user switch

NiceandGently:
this fixes https://github.com/mate-desktop/mate-session-manager/issues/76


NiceandGently:
please cherry-pick to 1.8

PR #80, opened 2015-04-01
ckeck for running DM for user switching

NiceandGently:
fixes https://github.com/mate-desktop/mate-session-manager/issues/76

taken from linuxmint/cinnamon-session@55538a5


monsta:
@stefano-k: works fine for me in Debian Jessie

monsta:
@stefano-k: hmm, 1.10 is tagged already - and without this PR? :(

PR #81, opened 2015-04-01
1.8: ckeck for running DM for user switching

NiceandGently:
fixes mate-desktop#76

taken from linuxmint/cinnamon-session@55538a5


flexiondotorg:
Thanks for your contribution.

Issue #86, opened 2015-06-18
XDG_CURRENT_DESKTOP not set


cygwinports:
In MATE (tested on Fedora 21), `XDG_CURRENT_DESKTOP` is empty.  This can used by GIO to set desktop-specific default applications (e.g. so that pluma is the default text editor even when GNOME and gedit are installed):

http://standards.freedesktop.org/mime-apps-spec/mime-apps-spec-latest.html
https://git.gnome.org/browse/glib/tree/gio/gdesktopappinfo.c

This can be accomplished by adding the following line to `data/mate.desktop.in` (which is installed as `$datadir/xsessions/mate.desktop`):

    DesktopNames=MATE

PR #87, opened 2015-06-19
set XDG_CURRENT_DESKTOP env to MATE

NiceandGently:
fixes https://github.com/mate-desktop/mate-session-manager/issues/86


flexiondotorg:
Closes #86

PR #93, opened 2015-08-03
RFC: Add option to disable/enable overlay scrolling

infirit:
This is a proof of concept and RFC only.

The recent Gtk3 scrolled window (bar) changes called overlay scrolling is really annoying me. And when @NiceandGently told me it could be disabled I was happy and started experimenting.

Some things still to add and discuss.

 * It should listen to gsettings changes and set the env variable appropriately
 * It overwrites the env variable regardless of what was there before (perhaps set by the user).
 * It has to run early in session start. So my thought was it needs to be in mate-session.
 * Anything else I forgot?

The first is an implementation detail which will be sorted before commit. But the second and third I am not sure.

On the second, should we blindly overwrite the env variable? Maybe it should be a "force old scrollbar" and ignore everything if this is disabled?

On the third, it needs to run early so for simplicity's sake only I it shoved it in msm-gnome. But the general question is should it go in mate-session-manager at all?

ps: The code relies on https://github.com/infirit/mate-desktop/commit/377eed97b112eced8be08a9ce7a2db25e191325d

### DIFF
--- a/mate-session/msm-gnome.c
+++ b/mate-session/msm-gnome.c
@@ -43,6 +43,8 @@
 
 #define GSM_SCHEMA "org.mate.session"
 #define GSM_GNOME_COMPAT_STARTUP_KEY "gnome-compat-startup"
+#define MATE_INTERFACE_SCHEMA "org.mate.interface"
+#define MATE_GTK_OVERLAY_SCROLL "gtk-overlay-scrolling"
 
 #define GNOME_KEYRING_DAEMON "gnome-keyring-daemon"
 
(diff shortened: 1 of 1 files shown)


NiceandGently:
First, a big thank you to @leigh123linux who got this idea for cinnamon in fedora.
On first
great if it is possible
On second
Personaly i like the new srollbars behaviour, but i think most mate users are more conservative and this behaviour causes also some issues in applications, ie. entries at the bottom of a window aren't selectable in some applications.
Also, not all distros using gtk+-3.16, so changing that at this point is OK for me.
I mean what is bad to say to a user/admin this is our decision and we have reasons to do that, you can always change the behaviour via gsettings or with a build override.
On third
I would prefer a more general place because !gtk+=gnome  

+1 for changing that in 1.10 branch

monsta:
Good idea. These appearing/disappearing scrollbars are annoying as hell - especially when you try to select the last package in Synaptic and the scrollbar is horizontal :confused:

stefano-k:
I agree with this proposal

flexiondotorg:
+1

NiceandGently:
Hey, team has voted for a go!
https://github.com/infirit/mate-session-manager/commit/00fe326fdd3796d38393822b4241ec16cc574a33
is missing :)

PR #96, opened 2015-09-17
Add support for an optional dock, such as Plank.

flexiondotorg:
The `required-components-list` adds an entry for `dock` which is null by default so that behaviour is unchanged. This change makes it possible to better integrate docks into the MATE session rather than manipulating autostart files.

### DIFF
--- a/data/org.mate.session.gschema.xml.in
+++ b/data/org.mate.session.gschema.xml.in
@@ -32,7 +32,7 @@
       <description>List of applications that are part of the default session.</description>
     </key>
     <key name="required-components-list" type="as">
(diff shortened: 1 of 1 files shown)


flexiondotorg:
Recommend this is cherry-picked for 1.10 as it doesn't change the default behaviour.

flexiondotorg:
@clefebvre, @stefano-k - What are your thoughts on this?

clefebvre:
Looks good. Should be 1.12 though as it's a new feature.

flexiondotorg:
Yeah, on reflection I agree about including it for 1.12.

PR #101, opened 2015-10-15
Make QT5 apps follow the GTK style

clefebvre:
You can use dropbox systray to test:

- Without this, the menu uses the QT style (white with blue highlight)
- With this, the menu looks like according to your GTK style

### DIFF
--- a/mate-session/main.c
+++ b/mate-session/main.c
@@ -631,6 +631,9 @@ int main(int argc, char** argv)
 	 */
 	gsm_util_setenv("MATE_DESKTOP_SESSION_ID", "this-is-deprecated");
 
+	/* Make QT5 apps follow the GTK style */
+	gsm_util_setenv ("QT_STYLE_OVERRIDE", "gtk");
+
 	/*
 	 * Make sure gsettings is set up correctly.  If not, then bail.
 	 */


raveit65:
real nice, works well with dropbox applet menu.
+1

flexiondotorg:
This variable doesn't need to be set in the session manager. I've been setting this via Xsession.d. It is also worth noting some Qt apps will segfault with this variable set.

monsta:
Yes, ```/etc/X11/Xsession.d/``` might be a better place for it. Current solution doesn't affect Qt5 apps running as root...

@flexiondotorg: maybe this better be done at distro level rather than in all GTK+ desktop environments out there...

flexiondotorg:
@monsta Yes, I would prefer this to be directly user configurable in case of regression.

clefebvre:
Ok, I'll fix it in Mint specifically then.

lukefromdc:
How do you go about doing this in /etc/X11/Xsession.d? I NEED this functionality as otherwise I have to manually change every panel launcher and menu entry for QT5 apps to force the theme.

flexiondotorg:
@lukefromdc You'll find how I do it in Ubuntu MATE here - https://code.launchpad.net/~ubuntu-mate-dev/ubuntu-mate/ubuntu-mate-settings

lukefromdc:
Thanks! That worked for me, and I don't have any of the apps that segfault over it.

PR #102, opened 2015-10-15
Only wait 1 second max on logout/shutdown

clefebvre:
Ported from https://github.com/linuxmint/cinnamon-session/commit/73048abbfe05f882fed31365a1fe5b5733b6b7f6


raveit65:
seems to working
+1

Issue #103, opened 2015-10-21
[1.11] mate-session-save tool doesn't work right after upgrade from 1.10


monsta:
Steps to reproduce:

1. Have MATE 1.10 installed.
2. Upgrade ```mate-desktop``` to 1.11 (see https://github.com/mate-desktop/debian-packages/commit/aca038f1235bb2d5801ee56bd69fb277cca0d376).
3. Upgrade ```mate-session-manager``` to 1.11 (also it's better to upgrade ```mate-settings-daemon``` to 1.11 at the same time, see https://github.com/mate-desktop/debian-packages/commit/f106be84565b5baae5d4e21c7a0dc9754574b9ea).
4. Try running ```mate-session-save``` with ```--logout-dialog``` or ```--shutdown-dialog``` or other arguments.
It will fail with an error message like this:
```
** (mate-session-save:2951): WARNING **: Failed to call logout: The name org.gnome.SessionManager was not provided by any .service files
```

It doesn't work as it's 1.11 version of the tool, so it tries to use ```org.gnome.SessionManager``` (after https://github.com/mate-desktop/mate-session-manager/commit/b01c41715ddad682c2e0d2a5eac8fa0deb91c5e8) - but the currently running session still uses ```org.mate.SessionManager```.

This affects the apps or applets that rely on ```mate-session-save``` to provide logout/shutdown ability (e.g. mintMenu): you can't log out or shut down the system. Of course the issue will be gone on the *next* session run, but... :smile:

This doesn't affect the built-in ```menu``` and ```menu-bar``` applets. They use panel's own code which still uses ```org.mate.SessionManager``` at the time. (Well, if you upgrade the panel to 1.11 as well, then restart it via ```killall mate-panel```, then it will cease to work; but would you do that? :smile:)

raveit65:
But a session restart after update of all affected packages should help?

monsta:
Yes, or some other solution.

@clefebvre @flexiondotorg @willysr @posophe @obache
So, we really need to solve this before 1.12 release - either upstream or in all distros...
Thoughts guys?

willysr:
upstream is preferred

monsta:
I was actually wondering *how* we'd solve it :smile:

raveit65:
dowstream maintainers need to do the update in one rush and marked it as 'restart is needed', if they are brave enough to update in a running release cycle ;)

monsta:
[...]

PR #104, opened 2015-10-21
mate-session-save: try org.mate.SessionManager if org.gnome fails

monsta:
fixes https://github.com/mate-desktop/mate-session-manager/issues/103


clefebvre:
+1

flexiondotorg:
+1

raveit65:
if it works for you +1

phattl:
I don't understand, how do I use org.mate.SessionManager?

clefebvre:
It's not used anymore, this just fixed a particular case during an upgrade.

PR #109, opened 2015-11-11
get rid of dependency on libmate-desktop

monsta:
this introduces a bit of code duplication but makes m-s-m independent
of any MATE packages which rely upon GTK+, and so makes it possible
to migrate m-s-m to GTK+3 right away.

of course, this change might be reverted in the future (when all of MATE
becomes GTK+3 only).

@clefebvre @flexiondotorg @raveit65 @willysr @posophe @obache 
In distros, build and runtime deps on ```libmate-desktop``` can be dropped after this.
Runtime dep on ```mate-desktop```'s schemas package (in Debian/Ubuntu/Mint it's ```mate-desktop-common```) is still needed of course.

Also we can hold it until 1.13 dev phase starts, or accept it for 1.12.1 - I'm fine with both options :smile:


raveit65:
I want to test it some days more because it is a sensible mate application.
If all works well i'm +1 for 1.12.1 than i can switch to gtk3 in fedora rawhide.

flexiondotorg:
Code duplication is minimal. +1 from me for 1.12.1.

posophe:
Well, if it's temporal. Why not

PR #110, opened 2015-11-14
compile fix for upower 0.99 series

oleid:
upower 0.99 has lost support for hibernate and suspend. This leads to undefined
references to up_client_get_can_suspend  and up_client_get_can_hibernate on my
system. This patch removes any call to those functions for upower >= 0.99.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,13 +146,18 @@ AC_ARG_ENABLE(upower,
               enable_upower=$enableval,
               enable_upower=no)
 if test "x$enable_upower" = "xyes"; then
-    PKG_CHECK_MODULES(UPOWER, upower-glib >= $UPOWER_REQUIRED, has_upower=yes, has_upower=no)
+    PKG_CHECK_MODULES([UPOWER], [upower-glib >= $UPOWER_REQUIRED], has_upower=yes, has_upower=no)
 
     if test "x$has_upower" = "xyes"; then
         AC_DEFINE(HAVE_UPOWER, 1, [upower support])
         AC_SUBST(UPOWER_CFLAGS)
         AC_SUBST(UPOWER_LIBS)
     fi
+    PKG_CHECK_MODULES([UPOWER_HIBERNATE], [upower-glib < 0.99], has_upower_hibernate_suspend=yes, has_upower_hibernate_suspend=no)
+    if test "x$has_upower_hibernate_suspend" = "xyes"; then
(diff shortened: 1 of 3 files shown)


clefebvre:
strange... msm compiled in both LMDE and Mint.. with LMDE using the newer upower.

oleid:
Probably it's related to systemd as well. I *guess* LMDE and Mint don't ship with systemd and thus, they use an upower version with this functionality.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=752425
Both,
```
$ strings /usr/lib/libupower-glib.so | grep hibernate
$ strings /usr/lib/libupower-glib.so | grep suspend 
```
don't return anything on my Gentoo/systemd and on my ArchLinux system.

raveit65:
same in fedora
```
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so | grep hibernate
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so.3 | grep hibernate
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so.3.0.1 | grep hibernate
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so.3.0.1 | grep suspend
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so.3 | grep suspend
[rave@mother ~]$ strings /usr/lib64/libupower-glib.so | grep suspend
```
don't return anything.
But i didn't get any undefined references during compile.

oleid:
@raveit65 Are you sure upower is enabled during configure? It seems to be disabled in the spec file in rawhide:

 ftp://rpmfind.net/linux/fedora/linux/development/rawhide/source/SRPMS/m/mate-session-manager-1.12.0-1.fc24.src.rpm

raveit65:
well, you're are right...to much packages to maintain :)
Here in fedora i don't need upower because all is handled by systemd-login itself.
So i never stumbled in this build issue.
Are you shure you need upower with systemd on gentoo?

tarakbumba:
I can confirm this issue and PR fixes that. Today i tried to rebuild mate-session-manager with upower enabled on my Mageia machine. Mageia Cauldron provides upower-glib = 0.99.3 and build fails. This PR fixes build.

monsta:
Oops. Good catch, thanks.

Speaking of LMDE 2, we have systemd support set to "auto" (no build options, just autodetect systemd lib) and don't use ```--enable-upower```, so we missed this. No one tried to build it with support for upower 0.99 which basically does nothing here... :smile:

monsta:
Fix is also ported to 1.12 branch and released in version 1.12.2 today.

PR #114, opened 2015-12-09
Enforce screen locking before requesting switch to a different user.

sunweaver:
One of my customers noticed that there is a security issue with user switching via the logout dialog.

The session does not get locked when user switching is requested via the logout dialog. However, this should be enforced.

I am not sure if the proposed patch is a generic way of requesting a screen lock. However, the approach works well on mate-session-manager as found in Debian jessie.

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -3178,6 +3178,19 @@ logout_dialog_response (GsmLogoutDialog *logout_dialog,
         case GTK_RESPONSE_DELETE_EVENT:
                 break;
         case GSM_LOGOUT_RESPONSE_SWITCH_USER:
+
+                ; /* place an empty statement between label above and declaration below... */
+
+                /* Lock screen before requesting user switch
+                 */
+                GError *error;
+                error = NULL;
+                g_spawn_command_line_async ("mate-screensaver-command --lock", &error);
+                if (error != NULL) {
+                    g_warning ("Couldn't lock screen: %s", error->message);
+                    g_error_free (error);
+                }
+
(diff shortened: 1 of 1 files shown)


clefebvre:
There are other ways to switch users, some of them not being shipped by MATE, so it's usually the responsibility of the DM to lock the session. Either by asking the session to lock itself, or by locking screensavers directly. 

See here for instance: 

https://github.com/linuxmint/mdm/blob/master/gui/mdmflexiserver.c#L206

If your DM does not lock your session, make sure to report it as a bug on the DM as well.

Now, with that said, it should be quite harmless to accept this PR. It shouldn't create issues with MDM and it should help a little with DMs which don't support mate-screensaver (at least in the cases where the user switches users via mate-screensaver).

clefebvre:
Btw, it might be better to make the call in manager_switch_user() directly.

raveit65:
Well, mdm is only use by linuxmint.
In fedora with lightdm i can always switch between 2 user accounts wit alt-f1/f2 until the screensaver lock gets in.
Fedora use always another ttyl for a second/third.....user.
This might be different to debian systems.
This is really a long outstanding issue and it is already reported somewhere here at github.

joakim-tjernlund:
Could this be enforced when issuing a suspend as well?
Now suspend does not so one have to wake up the suspend twice unless I lock screen manually
first.

joakim-tjernlund:
Forget the suspend issue, that was due to something else.

vkareh:
Both GDM and LightDM handle this correctly. MDM is no longer an active project. Closing.

PR #121, opened 2016-05-18
require GTK+ >= 3.14, drop GTK+2 code and --with-gtk build option

monsta:
@flexiondotorg @clefebvre @raveit65 @posophe @XRevan86 @willysr @obache @NP-Hardass

So, now we move m-s-m to GTK+3 as well:

- minimum GTK+ version is set to 3.14
- GTK+2 code and deprecated GTK+3 code are dropped
- ```--with-gtk``` build option is dropped as well
- build-dep on GTK+ dev package is changed (that's what you need to do in your distros)

### DIFF
--- a/capplet/gsm-app-dialog.c
+++ b/capplet/gsm-app-dialog.c
@@ -254,10 +254,6 @@ gsm_app_dialog_constructor (GType                  type,
 
         setup_dialog (dialog);
 
-#if !GTK_CHECK_VERSION (3, 14, 0)
-        gtk_widget_show_all (GTK_WIDGET (dialog));
-#endif
-
         return G_OBJECT (dialog);
 }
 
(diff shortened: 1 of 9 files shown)


willysr:
+1 from me

flexiondotorg:
Excellent work @monsta

raveit65:
Looks like people thinks this is a voting :smile: 
...i think @monsta wants that people test the commit.
...building.......installation.......runtime :wink:

monsta:
It's also a notification for distro maintainers so they won't get a surprise when 1.16 is released :smile:

lukefromdc:
Built, installed, killed Xorg and logged right back in with this, no problems at all. Running with GTK3 for past 14 months anyway but figured I'd better check this build for any errors.

PR #129, opened 2016-10-19
Also look for gdm-binary process

gapan:
In some systems the gdm binary is actually called "gdm-binary", so we
need to check for that too.

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -1086,7 +1086,7 @@ manager_switch_user (GsmManager *manager)
                         g_error_free (error);
                 }
         }
-        else if (process_is_running("gdm") || process_is_running("gdm3")) {
+        else if (process_is_running("gdm") || process_is_running("gdm3") || process_is_running("gdm-binary")) {
                 /* GDM */
                 command = g_strdup_printf ("%s %s",
                                                GDM_FLEXISERVER_COMMAND,


monsta:
Wasn't that only in very old gdm (about 2.20.x) from gnome 2?

gapan:
Maybe. So?

monsta:
Hmm, they actually kept that stuff until 2013:
https://git.gnome.org/browse/gdm/commit/?id=b6c11aa4626f98088fdacf60cd69d6cff22b14bc

That's gdm 3.8.0. Which distro still ships that?

gapan:
Well, mine does. Does in hurt in any way to add this check?

monsta:
Wow... you have gdm 2.20.11 in Salix? Alright, I'll merge this :smile:

gapan:
Yep. It does what we want it to do without introducing dependencies that we
don't want. If someone finds a bug, we fix it. If we move to something else
in the future it will be something entirely different and probably not gdm3.

monsta:
BTW, you can check [MDM](https://github.com/linuxmint/mdm), a fork of GDM 2.20.x, for some possible bugfixes to pick. Of course, the code isn't the same now, but I remember there were some fixes during its development since 2012.

gapan:
Yeah, thanks. I remember having some trouble with it, but don't remember what exactly. Will try it again.

PR #130, opened 2016-12-21
Allow TreeView to expand vertically

jconroy77:
Resizing the Startup Applications Preferences window vertically does not result in the startup programs list expanding to fill the available space (though horizontal window expansion was filled automatically).

This change was all that appeared necessary to get the startup program listing to expand vertically to fill the resized window.

Tested with Mate 1.16 on CentOS 7.

### DIFF
--- a/data/session-properties.ui
+++ b/data/session-properties.ui
@@ -44,6 +44,7 @@
                     <property name="height_request">210</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
+                    <property name="vexpand">True</property>
                   </object>
                 </child>
               </object>


raveit65:
Well, this works but this commit looks like to fix the same.
https://github.com/mate-desktop/mate-session-manager/commit/f15be61545192568dfcb438d3452d637f08001ab
@monsta ???

monsta:
Yes, I've already made a (different) change, and it's fixed. You can patch 1.16 to test it.

jconroy77:
Pulled patch and tested.  Previous commit does fix the issue for me.  Closing this Pull Request.

raveit65:
just test it in cento7 VM and it works.
@jconroy77 
any ways, thanks for the PR, i will update epel7 build soon.

Issue #131, opened 2017-01-03
Open TCP ports by default on Arch Linux


AndyRandom:
The MATE Session Handler (/usr/bin/mate-session) default configuration is to listen on a TCP V4 and TCP V6 port.
On all interfaces !
**This is a total unnecessary security risk**.

Package version:
Arch Linux i386 (but very likely on all platforms). mate-session-manager 1.16.0-2 (mate mate-gtk3)

config files etc.
/etc/lightdm/lightdm.conf
[XDMCPServer]
#enabled=false

ps axu | grep -e "X[org]"
/usr/lib/xorg-server/Xorg :0 -seat seat0 -auth /run/lightdm/root/:0 -nolisten tcp vt7 -novtswitch

lsof -i -n -P
COMMAND PID USER FD TYPE DEVICE SIZE/OFF NODE NAME
mate-sess 899 me 13u IPv6 22760 0t0 TCP *:35757 (LISTEN)
mate-sess 899 me 14u IPv4 22761 0t0 TCP *:43365 (LISTEN)

Steps to reproduce:
startx

The software author is aware of the problem:
grep -A1 -ne "security reasons" mate-session-manager-1.16.0/mate-session/gsm-xsmp-server.c
566: * hosts, so for security reasons it would be best if ICE didn't
567- * even open any non-local sockets. So we use an internal ICElib

There are user complains since back from 2014:
https://bbs.archlinux.org/vietopic.php?id=182726
https://bbs.archlinux.de/viewtopic.php?id=25645
with ZERO REACTION ! (and some wrong advise...)
**Is this on purpure to increase the attack area of a default Mate installation ?**
I couldn't find any documentation, nowhere, how to stop that.

Any suggestions ?

lots0logs:
Could you check to see if you have `xtrans` installed? If not, try installing it and then restart your session. Does the issue persist?

AndyRandom:
I installed xtrans.

Name            : xtrans
Version         : 1.3.5-1
Description     : X transport library
Architecture    : any
URL             : http://xorg.freedesktop.org/
Licenses        : custom
Groups          : None
Provides        : None
Depends On      : None
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : None
Replaces        : None
Installed Size  : 307.00 KiB
Packager        : Andreas Radke <andyrtr@archlinux.org>
Build Date      : Wed 24 Sep 2014 05:37:51 NZST
[...]

PR #132, opened 2017-01-18
build: check stricter for presence of xtrans

monsta:
this should ensure that mate-session won't listen to TCP ports.
if needed, this check can be disabled using --without-xtrans option.

fixes https://github.com/mate-desktop/mate-session-manager/issues/131

adapted from:
https://git.gnome.org/browse/gnome-session/commit/?id=aa4c9d125012ae94154ce075d62342b63b41a7d9

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -243,12 +243,18 @@ fi
 AC_SUBST(EXECINFO_LIBS)
 
 dnl ====================================================================
-dnl Check for newish X interface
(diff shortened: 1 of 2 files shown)


lukefromdc:
I strongly concur with this, to the point that I am going to push my Debian build of it to my Archive site as a security update. I don't know if Debian was hitting this or not but this is something to take precautions rather than chances with. Build went fine, I then immediately restarted my session.

PR #133, opened 2017-01-18
Replace deprecated upower functions with ConsoleKit2 equivalents

joakim-tjernlund:
Forward port of Infrit's CK2 patch 
This requires ConsoleKit2 version 0.9.2.

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -38,10 +38,6 @@
 #include <dbus/dbus-glib.h>
 #include <dbus/dbus-glib-lowlevel.h>
 
-#ifdef HAVE_UPOWER
-#include <upower.h>
-#endif
-
 #include <gtk/gtk.h> /* for logout dialog */
 #include <gio/gio.h> /* for gsettings */
 
@@ -150,10 +146,6 @@ struct GsmManagerPrivate
 
         DBusGProxy             *bus_proxy;
         DBusGConnection        *connection;
-#ifdef HAVE_UPOWER
-        /* Interface with other parts of the system */
-        UpClient               *up_client;
-#endif
 };
 
 enum {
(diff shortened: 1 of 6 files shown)


mate-session/gsm-manager.c
```diff
@@ -1178,26 +1170,16 @@ manager_attempt_hibernate (GsmManager *manager)
 
                 gsm_systemd_attempt_hibernate (systemd);
         }
-#endif
-#if defined(HAVE_SYSTEMD) && defined(HAVE_UPOWER_HIBERNATE_SUSPEND)
-        else {
-#endif
-#ifdef HAVE_UPOWER_HIBERNATE_SUSPEND
-        gboolean can_hibernate = up_client_get_can_hibernate (manager->priv->up_client);
+#else
```
monsta: I might have missed something, but this looks like a transition to compile-time only choice. Since other parts of this file have both compile-time and runtime choice between systemd and consolekit, I suggest doing the same here.

The same applies to the suspend code changes a few lines below.

mate-session/gsm-manager.c
```diff
@@ -3376,14 +3340,6 @@ gsm_manager_can_shutdown (GsmManager *manager,
 #ifdef HAVE_SYSTEMD
         GsmSystemd *systemd;
 #endif
-        gboolean can_suspend = FALSE;
```
monsta: How did I miss this... removing this code block breaks build with systemd enabled :open_mouth: 

Well, the logic was flawed anyway. We were asking upower if we can suspend/hibernate when we could ask systemd instead. I'll correct it.

monsta:
Merged with corrections for mentioned issues - https://github.com/mate-desktop/mate-session-manager/commit/9923739fe6f161a05078340593bf6a91db13d9be.

PR #138, opened 2017-05-07
Update environment for dbus and systemd user-session

besser82:
Same as for [cinnamon-session](https://github.com/linuxmint/cinnamon-session/pull/91).
Backported from recent gnome-session.

### DIFF
--- a/mate-session/gsm-util.c
+++ b/mate-session/gsm-util.c
@@ -551,21 +551,140 @@ gsm_util_export_activation_environment (GError     **error)
         return environment_updated;
 }
 
+#ifdef HAVE_SYSTEMD
+gboolean
+gsm_util_export_user_environment (GError     **error)
+{
+        GDBusConnection *connection;
+        gboolean         environment_updated = FALSE;
+        char           **entries;
+        int              i = 0;
+        GVariantBuilder  builder;
+        GRegex          *regex;
+        GVariant        *reply;
+        GError          *bus_error = NULL;
+
+        connection = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, error);
+
+        if (connection == NULL) {
+                return FALSE;
+        }
+
+        regex = g_regex_new ("^[a-zA-Z_][a-zA-Z0-9_]*=([[:blank:]]|[^[:cntrl:]])*$", G_REGEX_OPTIMIZE, 0, error);
+
+        if (regex == NULL) {
+                return FALSE;
+        }
+
+        g_variant_builder_init (&builder, G_VARIANT_TYPE ("as"));
+        for (entries = g_get_environ (); entries[i] != NULL; i++) {
+                const char *entry = entries[i];
+
+                if (!g_utf8_validate (entry, -1, NULL))
+                    continue;
+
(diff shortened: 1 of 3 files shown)


raveit65:
This fixes an annoying issue with some applets ,ie. clock applet.
Without the PR those applets use system-default locale for context menu and not the user language.
A patch for 1.16 you will find here http://pkgs.fedoraproject.org/cgit/rpms/mate-session-manager.git/tree/mate-session-manager-1.16-upd-env-dbus-systemd.patch?h=f25
This fixes the user locale problem with f24/25/26 for me, just tested.

raveit65:
merged https://github.com/mate-desktop/mate-session-manager/commit/8f84934b86f04eebb9d7aebf32321925e1764958
Thank you

flexiondotorg:
@besser82 This commit causes a regression on Ubuntu MATE 17.10 daily. I'm exporting `$GTK_MODULES` via [/etc/X11/Xsession.d/96vala-panel-appmenu](https://github.com/rilian-la-te/vala-project-packages/blob/master/debian/vala-panel-appmenu/debian/96vala-panel-appmenu) which is sourced after `95dbus_update-activation-env`, here's the contents:

```bash
# vim:set ft=sh sw=2 sts=2 et:

if [ -n "$DBUS_SESSION_BUS_ADDRESS" ] && \
    [ -x "/usr/bin/dbus-update-activation-environment" ]; then
  # subshell so we can unset environment variables
  (
    # unset login-session-specifics
    unset XDG_SEAT
    unset XDG_SESSION_ID
    unset XDG_VTNR

    # tell dbus-daemon --session (and systemd --user, if running)
    # to put the Xsession's environment in activated services'
    # environments
    dbus-update-activation-environment --verbose --systemd --all
  )
fi
```
Using `mate-session-manager` 1.18.0 (*which does not include your patches*) I have to use the order above so that Indicators function when `unity-gtk-module` is exported and `mate-applet-appmenu` (*from [vala-panel-appmenu](https://github.com/rilian-la-te/vala-panel-appmenu)*) is also active in a panel. If I source `9?vala-panel-appmenu` before `95dbus_update-activation-env` then Indicators do not get displayed at all when `mate-applet-appmenu` and `mate-indicator-applet` are both running. I ran in to this issue a few days ago and raised the [mate-indicator-applet doesn't show any indicators](https://github.com/rilian-la-te/vala-panel-appmenu/issues/114) bug upstream, then closed it when I found the solution I describe above.

When using `mate-session-manager 1.18.1` (*which does include your patches*) regardless of where or how I export `GTK_MODULES="unity-gtk-module"` and `UBUNTU_MENUPROXY=1` the Indicators never show when `mate-applet-appmenu` and `mate-indicator-applet` are running.

Any idea what might be going on here? 

Tagging @monsta and @rilian-la-te as they may have some insight too.

besser82:
@flexiondotorg Well, the DM should not set nor alter the environment at all…  All user-side environment should be setup in `~/.bashrc`, system-wide env in some file located in `/etc/profile.d/`.
If $ENV is setup properly, the way I described, then it should work perfectly with my patch; it works the same way `GNOME-Session` does.

PR #152, opened 2018-01-27
require GTK+ 3.22 and GLib 2.50

monsta:
and drop additional checks for min/max GLib versions, it's not needed

### DIFF
--- a/capplet/main.c
+++ b/capplet/main.c
@@ -38,7 +38,6 @@ static GOptionEntry options[] = {
 
 static void dialog_response(GsmPropertiesDialog* dialog, guint response_id, gpointer data)
 {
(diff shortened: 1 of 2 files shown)


monsta:
I don't think we need these macros.

https://developer.gnome.org/glib/stable/glib-Version-Information.html#GLIB-VERSION-MIN-REQUIRED:CAPS

As I see, they just add more warnings. We already get the deprecations warnings anyway.

raveit65:
Ok , so the correct term is GLIB_VERSION_MIN_REQUIRED and GLIB_VERSION_MAX_ALLOWED
I have the feeling this was mixed in same packages.
I agree with you, seeing the normal deprecation warning is enough and annoying too.

PR #163, opened 2018-07-06
Strip outer single quotes from string value returned by gsettings in …

sunweaver:
…mate-wm script.

 See Debian bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889907


raveit65:
Adding the `--replace` option to marco `opt` doesn't have any effect.
So i am still wondering if this script will ever work without replacing an running WM.

Issue #157, opened 2018-04-15
mate-session --version returns variable program name/path


ewangi:
I'm not sure if this is a bug or a feature, but the output from **mate-session --version** produces a variable program name/path depending on how it is called. This may result in confusion for those applications which programmatically parse the output (e.g. those that look for ^mate-session to start the line).  

#### Expected behaviour
--version returns "mate-session 1.18.2" (with a variable version number), no matter how it is called.

#### Actual behaviour
--version returns different program name depending on how called:

```
$ mate-session --version
mate-session 1.18.2

$ /usr/bin/mate-session --version
/usr/bin/mate-session 1.18.2
```

Running MATE 1.18.2 on Linux Mint 18.3.

yetist:
I think it's a bug, other applications output like this:
```
$gimp --version
GNU 图像处理程序 版本 2.10.8
$/usr/bin/gimp --version
GNU 图像处理程序 版本 2.10.8
$chromium --version
Chromium 70.0.3538.77 Arch Linux
$/usr/bin/chromium --version
Chromium 70.0.3538.77 Arch Linux
```

yetist:
This problem has been fixed and the code has been merged into the master branch. Can we close the issue?

sc0w:
@yetist yes, of course, thanks!

Fixed in https://github.com/mate-desktop/mate-session-manager/pull/168 with https://github.com/mate-desktop/mate-session-manager/commit/7df5759ff520c502314205bc2887f6467058fbfc

@yetist in your future commits, if you like, you can say into the commit message "Fixes #numberissue" and the issue automatically will be close if the PR is merged

PR #168, opened 2018-11-13
Fixed bug #157

yetist:
Run with `--version`, only show the application name and version, not show
the path of program.


monsta:
Not sure if everyone is okay with the commit message - it could be reworded to mention the actual issue in the first line

raveit65:
Sorry, a link doesn't belongs in commit header.
The link should be in commit body.
Please reword.

PR #169, opened 2018-11-26
gsm-manager: avoid deprecated 'g_settings_list_schemas'

sc0w:


sc0w:
Now, seems there are no more deprecated functions in mate-session-manager :)

monsta:
Achievement unlocked! :smile:

yetist:
Good, next should migrat dbus-glib to gdbus.

PR #170, opened 2018-11-27
Migrate from dbus-glib to gdbus

yetist:
Please review the code, thank you.


lukefromdc:
Should I merge this? Nobody has done anything further with it in a week

raveit65:
I dunno, i don't know anything about dbus, so i would prefer that @monsta take a look over it.
But if you think it is OK than go ahead.

lukefromdc:
OK, that looks like we have a problem here

yetist:
Sorry, I made a mistake here.

monsta:
Squashed + merged

PR #171, opened 2018-11-27
Migrate from dbus-glib to gdbus

yetist:
Please review, thank you.

### DIFF
--- a/mate-session/test-client-dbus.c
+++ b/mate-session/test-client-dbus.c
@@ -95,7 +95,7 @@ on_client_query_end_session (GDBusProxy     *proxy,
                                       -1,
                                       NULL,
                                       &error);
-        if (error != NULL) {
+        if (res == NULL) {
                 g_warning ("Failed to respond to EndSession: %s", error->message);
                 g_error_free (error);
         } else {
@@ -166,7 +166,7 @@ register_client (void)
                                       -1,
                                       NULL,
                                       &error);
-        if (error != NULL) {
+        if (res == NULL) {
                 g_warning ("Failed to register client: %s", error->message);
                 g_error_free (error);
                 return FALSE;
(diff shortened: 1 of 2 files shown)


lukefromdc:
Is this a duplicate of https://github.com/mate-desktop/mate-session-manager/pull/170 or should that PR be closed in favor of this one?

yetist:
Both #170 and #171 are migrate dbus-glib to gdbus, but different files are changed.
Maybe I should merge them into one PR?

sc0w:
@yetist It's your decision, but if you like to split, I think the files affected need to mentioned in the commit message and the title of PR

raveit65:
Only rewording the commit and adding the affected files to commit body is needed from my point of view.
In both PR's. Please.....

lukefromdc:
I do not know enough about GError to know whether  g_error_free only needs to be run when the GError variable has been set to a non-NULL value or if it always needs to  be run. Other than that memory management is probably OK. I don't know enough about either dbus system to spot any screwup in the use of gdbus, but it looks like a straight code replacement

lukefromdc:
Session saving also works fine, so no issues there

lukefromdc:
Should I merge this? Nobody has done anything further with it in a week

monsta:
I can't figure out how to use this test app...

mate-session/test-client-dbus.c
```diff
@@ -64,129 +58,138 @@ session_manager_connect (void)
...
+        res = g_dbus_proxy_call_sync (proxy,
+                                      "EndSessionResponse",
+                                      g_variant_new ("(bs)",
+                                                     is_ok,
+                                                     reason),
+                                      G_DBUS_CALL_FLAGS_NONE,
+                                      -1,
+                                      NULL,
+                                      &error);
+        if (error != NULL) {
```
monsta: Does `g_dbus_proxy_call_sync` set it to `NULL` on success, or just doesn't touch it in that case?
It's probably better to check the return value, like it's done in #170.

monsta: Same in two similar situations below

monsta:
Thanks. I'll squash + merge it as it's needed to have a working test app for #170.

PR #174, opened 2018-12-09
session-properties.ui: avoid deprecated GtkImage:stock

sc0w:
In this commit I added the property `translatable=yes` to the buttons, and I fixed the deprecated GtkImage:stock

### DIFF
--- a/data/session-properties.ui
+++ b/data/session-properties.ui
@@ -82,7 +82,7 @@
                 <property name="layout_style">start</property>
                 <child>
                   <object class="GtkButton" id="session_properties_add_button">
-                    <property name="label">_Add</property>
+                    <property name="label" translatable="yes">_Add</property>
                     <property name="visible">True</property>
                     <property name="can_focus">True</property>
                     <property name="receives_default">True</property>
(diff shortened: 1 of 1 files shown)


sc0w:
NOTE: this PR affects this icon, I have no idea how to show it

raveit65:
I never saw this button in this dialog.
But it is defined here https://github.com/mate-desktop/mate-session-manager/blob/master/capplet/gsm-properties-dialog.c#L44

sc0w:
Finally, I can see the button with gtkinspector, changing the property visible to true on the button, and the image is there, so, it must work

raveit65:
Yes, the visible flag was missing
```
              <object class="GtkButton" id="session_properties_save_button">
                <property name="visible">True</property>
                <property name="can_focus">True</property>
                <property name="receives_default">True</property>
                <child>
```

raveit65:
But i don't see any application in `~/.config/mate-session/saved-session/` if i click on the button.

sc0w:
lol, on click this function is called:

```
static void
on_save_session_clicked (GtkWidget           *widget,
                         GsmPropertiesDialog *dialog)
{
        g_debug ("Session saving is not implemented yet!");
}
```

It seems it is hidden intentionally

raveit65:
Seems ready to go?

PR #175, opened 2018-12-10
mate-session-save: show icon in error dialog

sc0w:


lukefromdc:
Not sure how to test this, but I don't see anything in the code that should create a problem

Issue #127, opened 2016-09-12
Add a delay setting to Startup Manager


Rahtgaz:
It would be useful to have a delay setting when adding new applications to mate-session-properties. Currently we have to manage delayed applications through user-made scripts and with the use of the `sleep` command.

However the number of startup applications that require delayed execution can be substantial on some systems, particularly if they depend on ordered execution or require a fully loaded environment. This is true of applications like Conky, for instance, or Tilda when using transparency with some window managers.
##

arxo:
Very useful idea, indeed. :) I've run into problems with startup applications in need of delays a number of times.

gitarra:
For your information, Linux Mint has this feature available.

See: https://ubuntu-mate.community/t/request-add-startup-delay-to-startup-applications/8876

danielw2904:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/modules/cs_startup.py

This is the mint implementation

danielw2904:
Specifically

`if "delay" in self.save_mask.contents:
            key_file.set_string(D_GROUP, "X-GNOME-Autostart-Delay", self.delay)`

yetist:
mate-session support `X-MATE-Autostart-Delay=10` in desktop file, is it enough?

vkareh:
Fixed by #178

PR #178, opened 2018-12-14
Add a delay setting to Startup Manager

yetist:
Fixed issue #127
- capplet/gsm-app-dialog.c
- capplet/gsm-app-dialog.h
- capplet/gsm-properties-dialog.c
- capplet/gsp-app.c
- capplet/gsp-app.h
- capplet/gsp-keyfile.h
- data/session-properties.ui

### DIFF
--- a/capplet/gsp-keyfile.h
+++ b/capplet/gsp-keyfile.h
@@ -34,6 +34,8 @@ extern "C" {
 #endif
 
 #define GSP_KEY_FILE_DESKTOP_KEY_AUTOSTART_ENABLED "X-MATE-Autostart-enabled"
+#define GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY       "X-GNOME-Autostart-Delay"
+#define GSP_KEY_FILE_DESKTOP_KEY_DELAY             "X-MATE-Autostart-Delay"
 
 void      gsp_key_file_populate        (GKeyFile *keyfile);
 
@@ -46,6 +48,9 @@ gboolean gsp_key_file_get_boolean      (GKeyFile       *keyfile,
                                         gboolean        default_value);
 #define gsp_key_file_get_string(key_file, key) \
          g_key_file_get_string (key_file, G_KEY_FILE_DESKTOP_GROUP, key, NULL)
+#define gsp_key_file_get_delay(key_file) \
+         (g_key_file_get_integer (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY, NULL) + \
+         g_key_file_get_integer (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_DELAY, NULL))
 #define gsp_key_file_get_locale_string(key_file, key) \
          g_key_file_get_locale_string(key_file, G_KEY_FILE_DESKTOP_GROUP, key, NULL, NULL)
 
@@ -53,6 +58,12 @@ gboolean gsp_key_file_get_boolean      (GKeyFile       *keyfile,
          g_key_file_set_boolean (key_file, G_KEY_FILE_DESKTOP_GROUP, key, value)
 #define gsp_key_file_set_string(key_file, key, value) \
          g_key_file_set_string (key_file, G_KEY_FILE_DESKTOP_GROUP, key, value)
+#define gsp_key_file_set_delay(key_file, value) \
+        do { \
+        if (g_key_file_has_key (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY, NULL)) \
+                g_key_file_remove_key (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY, NULL); \
+        g_key_file_set_integer(key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_DELAY, value); \
+        } while (0);
 void    gsp_key_file_set_locale_string (GKeyFile    *keyfile,
                                         const gchar *key,
                                         const gchar *value);
--- a/capplet/gsp-keyfile.h
+++ b/capplet/gsp-keyfile.h
@@ -34,7 +34,6 @@ extern "C" {
 #endif
 
 #define GSP_KEY_FILE_DESKTOP_KEY_AUTOSTART_ENABLED "X-MATE-Autostart-enabled"
-#define GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY       "X-GNOME-Autostart-Delay"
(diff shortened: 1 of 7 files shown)


capplet/gsp-keyfile.h
```diff
@@ -46,13 +48,22 @@ gboolean gsp_key_file_get_boolean      (GKeyFile       *keyfile,
                                         gboolean        default_value);
 #define gsp_key_file_get_string(key_file, key) \
          g_key_file_get_string (key_file, G_KEY_FILE_DESKTOP_GROUP, key, NULL)
+#define gsp_key_file_get_delay(key_file) \
+         (g_key_file_get_integer (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_GNOME_DELAY, NULL) + \
```
monsta: Wait... does this sum the values of two delay keys?

capplet/gsp-keyfile.h
```diff
@@ -46,13 +48,22 @@ gboolean gsp_key_file_get_boolean      (GKeyFile       *keyfile,
...
+         g_key_file_get_integer (key_file, G_KEY_FILE_DESKTOP_GROUP, GSP_KEY_FILE_DESKTOP_KEY_DELAY, NULL))
 #define gsp_key_file_get_locale_string(key_file, key) \
          g_key_file_get_locale_string(key_file, G_KEY_FILE_DESKTOP_GROUP, key, NULL, NULL)
 
 #define gsp_key_file_set_boolean(key_file, key, value) \
          g_key_file_set_boolean (key_file, G_KEY_FILE_DESKTOP_GROUP, key, value)
 #define gsp_key_file_set_string(key_file, key, value) \
          g_key_file_set_string (key_file, G_KEY_FILE_DESKTOP_GROUP, key, value)
+#define gsp_key_file_set_delay(key_file, value) \
```
monsta: Do we actually handle `X-GNOME-Autostart-Delay` (or any of their other keys) anywhere else? I can't find the code...
If not, then there's no need to remove that key.

lukefromdc:
Still works fine with the latest change

raveit65:
Hmm, it works only one time for me, tested with dropbox na-tray-applet + 10 secs delay.
The applet starts with chosen delay in fresh session but the delay setting is lost in mate-session-properties UI.
And some applications still ignore a delay setting here, eg. nm-applet and mate-volume-control applets.

lukefromdc:
I have been keeping this installed, and conky has been starting up just fine, time after time. Try a 9 second delay, maybe something wrong with a two decimal digit delay?

raveit65:
I tested it again with nm-applet +6 sec in a VM and the Networkmanager-applet starts without any delay in new session.
I noticed that after restarting mate-session-properties the setting for nm-applet is lost.
And i noticed this warning with closing mate-session-properties
```
[rave@f28 ~]$ mate-session-properties 

(mate-session-properties:2038): GLib-CRITICAL **: 10:02:11.602: Source ID 892 was not found when attempting to remove it
[rave@f28 ~]$ 
```

lukefromdc:
I was still unable to duplicate @raveit65's issue: added gedit with a 5 second delay, no issues on restarting the session, gedit came up with the 5 second delay. Opening and closing  mate-session-properties didn't throw that warning no matter what I did, and I could freely add and remove items without getting warnings. Question is thus what is different?

Again, I am testing on bare metal and running this at the moment as my main session manager

yetist:
The app in system path only changed delay time is ok now.

Known problems:
Some applications create or modify desktop files themselves, which will remove the delay value, like guake: 
https://github.com/Guake/guake/blob/fa6c0676ff271806b44531031ff869362c5f920a/guake/prefs.py#L328

lukefromdc:
I just tested this with network-manager-applet(though mine is a custom build), and the 5 second delay I tested it with was saved and worked.

raveit65:
@lukefromdc 
Why merging so fast?
I said loud an clear that it doesn't work here.
An i'd like to investigate why it doesn't work by me.
But the day doesn't have 96 hours and i have also other things to do.

raveit65:
After rebuilding m-s-m with last added commit from PR it seems to work with nm-applet and mate-volume-applet.
Not sure about caja-dropbox applet, seems this applet ignore the timing.
But i get randomly this error message if i am closing mate-session-properties, not all the time.
```
[rave@f28 ~]$ mate-session-properties 

(mate-session-properties:2038): GLib-CRITICAL **: 10:02:11.602: Source ID 892 was not found when attempting to remove it
```
But maybe this related to GDbus port.

raveit65:
I found the culprit with dropbox applet.
It seems like that dropbox always re-create  ~/.config/autostart/dropbox.desktop after session start.
So this a one shot setting for this applet.
And dropbox.desktop have `X-GNOME-Autostart-Delay=10` in fedora build from rpmfusion.
Looks like that this can cause the glib warning randomly , ihmo.

lukefromdc:
Sorry about that, I merged because I saw three approving reviews and didn't see further references to a problem after the last changes.  I was holding off seeing an issue, but then came the last round of changes and a third approving review.

 Question: if Dropbox's autostart is turned off entirely, does starting a new session turn it back on?  If so, we have a bigger issue in that dropbox(which I cannot test) and possibly some other apps are forcibly overwriting all user choices about session startup options.  We cannot fix an app that forcibly turns its own autostart back on  in our session manager.

 On the other hand, if it does NOT turn autostarting back on, we can turn it off, keep it off, and write a ciopy of the file under another name. A user should be able to do that right now by turning off the offending autostart job and copying its functions (possibly with a new name like "dropbox0") into a new autostart job. We could document that in the help and/or in the tooltip for the delay setting.

To automate this would be possible but would require either copying every last modified .desktop file to a new name (messy) or using a whitelist. We could have an option in the delay (maybe a "fix lost delay setting" button in the delay field) to copy the file.

I propose that such a fix for force-overwritten .desktop files be in a new PR, but if you want to revert this one for more work that is also OK by me. I can do the conky delay I need and use by my old script. A project like this is a team effort and what works for the whole project is what counts

raveit65:
> Question: if Dropbox's autostart is turned off entirely, does starting a new session turn it back on? If so, we have a bigger issue in that dropbox(which I cannot test) and possibly some other apps are forcibly overwriting all user choices about session startup options. We cannot fix an app that forcibly turns its own autostart back on in our session manager.

No, this is working. Our UI add simply `X-MATE-Autostart-enabled=false` to ~/.config/autostart/dropbox.desktop.
This will be respect because dropbox doesn't start any more.
But if i change eg. `X-GNOME-Autostart-Delay=10` to `X-GNOME-Autostart-Delay=5` by hand, than next start of dropbox will set it back to `10`.

> On the other hand, if it does NOT turn autostarting back on, we can turn it off, keep it off, and write a ciopy of the file under another name. A user should be able to do that right now by turning off the offending autostart job and copying its functions (possibly with a new name like "dropbox0") into a new autostart job. We could document that in the help and/or in the tooltip for the delay setting.

I think adding an extra autostart desktop file will generate a conflict with dropbox UI itself, because in dropbox UI is a setting for enable/disable autostart in session.

Maybe writing only a blacklist for applications which don't respect the delay which can be used for greyout the delay setting?

Any way, it works here with all Mate applications in auto start and most others, so it isn't a big problem.

Btw. no need to revert something in master.

PR #184, opened 2019-01-29
Backport restart user bus (second attempt)

### LINKED ISSUE #165: User processes are not terminated on logout
#### Problem description:
User processes are not terminated on logout. Apart from wasting resources, this is causing severe problems on re-logins, for example "cannot acquire bus" or even dconf database corruption.

#### Steps to reproduce:
1. Login
2. Logout
3. Press Alt+Ctrl+F1 to switch to vt1, and login there
4. Run `ps -ef` and see that most of the user processes are still running.

#### Expected behaviour
All other desktop environments properly terminate the user processes on logout.

#### MATE general version
mate 1.20.1

#### Package version
mate-desktop   1.20.1-2ubuntu1 

#### Linux Distribution
Ubuntu MATE 18.04.1

vkareh:
This finally seems to work. All process should terminate upon logout now.

How to test this:
* Log in to MATE
* Switch to a separate tty (`Ctrl+Alt+F2`)
* Run `ps aux | grep $USER` - there should be a million things
* Switch back to your graphical session (`Ctrl+Alt+F1` in Fedora, `Ctrl+Alt+F7` in Ubuntu, etc)
* Log out from MATE
* Switch back to tty
* Run `ps aux | grep $USER` - with this patch, there should not be any mate-related processes

Fixes #165

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -889,6 +889,51 @@ _client_stop (const char *id,
         return FALSE;
 }
 
+#ifdef HAVE_SYSTEMD
+static void
+maybe_restart_user_bus (GsmManager *manager)
+{
+        GsmSystemd *systemd;
+        GsmManagerPrivate *priv;
+        GDBusConnection *connection;
+
+        g_autoptr(GVariant) reply = NULL;
+        g_autoptr(GError) error = NULL;
+
+        priv = gsm_manager_get_instance_private (manager);
+        if (priv->dbus_disconnected)
+                return;
+
+        systemd = gsm_get_systemd ();
+
+        if (!gsm_systemd_is_last_session_for_user (systemd))
+                return;
+
+        connection = g_bus_get_sync (G_BUS_TYPE_SESSION, NULL, &error);
+
+        if (error != NULL) {
+                g_debug ("GsmManager: failed to connect to session bus: %s", error->message);
+                return;
+        }
+
+        reply = g_dbus_connection_call_sync (connection,
+                                             "org.freedesktop.systemd1",
+                                             "/org/freedesktop/systemd1",
+                                             "org.freedesktop.systemd1.Manager",
+                                             "TryRestartUnit",
+                                             g_variant_new ("(ss)", "dbus.service", "replace"),
+                                             NULL,
+                                             G_DBUS_CALL_FLAGS_NONE,
+                                             -1,
+                                             NULL,
+                                             &error);
+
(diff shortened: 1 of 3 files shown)


vkareh:
@lukefromdc - seems like partial success.

On my Ubuntu machine, the only process left with this patch was something related to gvfs (which is a _huge_ improvement from `master`, which left _a lot_ more things).

I just tried it on my Fedora laptop and the results were better than `master`, but not as good as what I got yesterday on Ubuntu, so I'm not entirely sure there...

vkareh:
@sc0w - thanks for bringing this up, I think Juha is right. From `man 3 sd_session_get_type`:
> **sd_session_get_type()** may be used to determine the type of the session identified by the specified session identifier. The returned string is one of "x11", "wayland", "tty", "mir" or "unspecified" and needs to be freed with the libc **free**(3) call after use.

I'll update that commit

sc0w:
Juha emailed me again:

```
Hi,

noticed also in that same commit
that when you go thru the sessions in for loop,
the first time you set

is_last_session = FALSE;

you do not need to go thru the remaining
sessions, if there are any left,
you can just break from the for loop.

This is not a bug, just caught my eye.

Regards,

Juha from Kuopio, Finland
```

vkareh:
Thanks again for relaying the information @sc0w. I added the `break`. I'll test this for a few days to make sure my commits don't cause any issues, but so far I've been using it without problems.

raveit65:
I found out that those 2 processes are started by lightdm/slickgreeter before i login to mate-session.
```
[root@f28 ~]# ps aux | grep rave
rave       881  0.0  0.2  87732  8096 ?        Ss   10:02   0:00 /usr/lib/systemd/systemd --user
rave       887  0.0  0.0 164560  3228 ?        S    10:02   0:00 (sd-pam)
```
So they are not started by mate-session and can be ignored for testing this PR.
After logout from mate session i found only 2 new user processes.
```
rave      2028  0.0  0.1 480960  5668 ?        Sl   10:07   0:00 /usr/libexec/geoclue-2.0/demos/agent
rave      2349  0.0  0.1  73024  4360 ?        Ss   10:07   0:00 /usr/bin/dbus-daemon --session --address=system
```
geoclue-2.0/demos/agent is in auto start but i have no idea for what this demo application is good for.
I have disable it on bare metal.
No idea what process `/usr/bin/dbus-daemon --session` did started why it is open. after logout.

Any way, much better that before.

PR #186, opened 2019-02-07
Fix make distcheck fail on debian.

yetist:
Please review and test it.


lukefromdc:
Works also without specifying the default WM with the latest commit

raveit65:
Translation source is updated and back ported to 1.20 branch.

PR #192, opened 2019-03-06
Guard against missing systemd

### LINKED ISSUE #190: mate-session-manager-1.22.0 FTBFS

vkareh:
Build fails without systemd support. I've added some guards against that scenario.

Fixes #190

### DIFF
--- a/mate-session/gsm-systemd.c
+++ b/mate-session/gsm-systemd.c
@@ -429,16 +429,31 @@ gsm_systemd_is_last_session_for_user (GsmSystemd *manager)
         gboolean is_last_session;
         int ret, i;
 
+#ifdef HAVE_SYSTEMD
         ret = sd_pid_get_session (getpid (), &session);
+#endif
 
-        if (ret != 0) {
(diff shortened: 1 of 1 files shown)


bunder2015:
works for me on gentoo building 1.22.0

vkareh:
@lukefromdc - no need to install, but could you confirm whether it builds for you if you configure with the `--without-systemd` flag? It builds for me on Fedora, tomorrow I'll test building it on Ubuntu

dhgutteridge:
FWIW, I've applied a similar patch to fix builds on non-Linux systems (NetBSD, SmartOS, et al.) via pkgsrc.org and am running mate-session-manager 1.22 under NetBSD without issue.

lukefromdc:
Build with  --without-systemd went fine too, no problem

vkareh:
Okay, I've gotten a lot of positive confirms from folks on IRC and GitHub that this patch works. It also builds well for me on both Ubuntu and Fedora, with or without the systemd flag.

I'm merging.

PR #194, opened 2019-03-08
Add ELOGIND support

vkareh:
Reuses HAVE_SYSTEMD #define to reduce #ifdef mess.

Fixes #146

PR #195, opened 2019-03-08
Add ELOGIND support

### LINKED ISSUE #146: mate-session compatibility with elogind
#### Expected behaviour

No [assertion](https://github.com/mate-desktop/mate-session-manager/blob/v1.18.1/mate-session/gsm-systemd.c#L567) in logs, doing similar code path what would happen if systemd-logind was running.

#### Actual behaviour

```
Aug 18 08:52:21 rotten-fruit mate-session[2823]: CRITICAL: gsm_systemd_set_session_idle: assertion 'session_path != NULL' failed
```

1. install [elogind](https://github.com/elogind/elogind)
2. suspend: `loginctl suspend`
3. wake up using keypress,  mouse, etc

#### MATE general version

1.18

#### Package version

mate-session-manager-1.18.0-1.x86_64

#### Linux Distribution

[PLD Linux](https://www.pld-linux.org/)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/48390415-mate-session-compatibility-with-elogind?utm_campaign=plugin&utm_content=tracker%2F581048&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F581048&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
Reuses HAVE_SYSTEMD #define to reduce #ifdef mess.

Fixes #146

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -110,6 +110,31 @@ AM_CONDITIONAL(HAVE_SYSTEMD, test "x$use_systemd" = "xyes")
 AC_SUBST(HAVE_SYSTEMD)
 
 dnl ====================================================================
+dnl Check for elogind
+dnl ====================================================================
+
+AC_ARG_WITH(elogind,
+            AS_HELP_STRING([--with-elogind],
+            [Use libelogind instead of libsystemd-login]),,
+            with_elogind=auto)
+
+use_elogind=no
+if test "x$with_elogind" != "xno"; then
+    PKG_CHECK_MODULES(LIBELOGIND,[libelogind], [use_elogind=yes],
+                      [PKG_CHECK_MODULES([LIBELOGIND], [libelogind],
+                      [use_elogind=yes], [use_elogind=no])])
+
+    if test "x$use_elogind" = "xyes"; then
+        AC_DEFINE([HAVE_ELOGIND], 1, [elogind support])
+        AC_DEFINE([HAVE_SYSTEMD], 1, [Define to 1 to reduce ifdefs since elogind is a drop-in replacement for systemd])
+        AC_SUBST(LIBELOGIND_CFLAGS)
+        AC_SUBST(LIBELOGIND_LIBS)
+    fi
+fi
+AM_CONDITIONAL(HAVE_ELOGIND, [test "x$use_elogind" = "xyes"])
+AC_SUBST(HAVE_ELOGIND)
+
+dnl ====================================================================
 dnl Check for XSync extension
 dnl ====================================================================
 
@@ -333,6 +358,7 @@ echo "
 
         Default WM:               ${with_default_wm}
         Systemd support:          ${use_systemd}
+        Elogind support:          ${use_elogind}
         IPv6 support:             ${have_full_ipv6}
         Backtrace support:        ${have_backtrace}
         XRender support:          ${have_xrender}
--- a/mate-session/Makefile.am
+++ b/mate-session/Makefile.am
@@ -7,6 +7,7 @@ noinst_PROGRAMS = 		\
 AM_CPPFLAGS =					\
 	$(MATE_SESSION_CFLAGS)		\
 	$(SYSTEMD_CFLAGS)			\
+	$(LIBELOGIND_CFLAGS)			\
 	$(DISABLE_DEPRECATED_CFLAGS)
 
 AM_CFLAGS = $(WARN_CFLAGS)
@@ -80,6 +81,7 @@ mate_session_LDADD =				\
 	$(XEXT_LIBS)				\
 	$(MATE_SESSION_LIBS)			\
 	$(SYSTEMD_LIBS)				\
+	$(LIBELOGIND_LIBS)			\
 	$(EXECINFO_LIBS)
 
 libgsmutil_la_SOURCES =				\


vkareh:
I don't use elogind and have no particular interest in it, but I'm happy to maintain this patch if it helps other people. I just need some consensus regarding versions, so that it doesn't break builds anywhere.

@mate-desktop/core-team - does anyone in our team uses elogind?

raveit65:
I don't know what it is and i don't use it.
And i am not really sure if it is a good idea to add support for something which nobody use from team
How do want to handle bug reports?
With support for consolekit2 we have already such a weird situation that nobody from team use it, but  users are thinking it is a supported feature.
Is it a good idea to promised users support for a feature, but nobody from team don't use it and can fix bugs?

joakim-tjernlund:
Removing the version req. is fine, i just added the latest stable elogind for Gentoo.
@raveit65 , elogind is a subset of systemd. The login/session stuff needed by apps like mate-session is there
but the whole services/init system is left out.

raveit65:
Ok, this for systemd haters ......
....... but my concern stays.
Who use it from the team and can give support for this code?

Personal i am very satisfied with systemd after all that years. Well, it isn't perfect, but which software is perfect nowadays.
I am honest, i don't like support code which i don't use.
But if someone like to do it , go ahead.

Btw. Why you don't patch your gentoo ebuild?

joakim-tjernlund:
I don't hate systemd, at work we have actually tried it twice but something always broke.
It is a big change to switch to systemd and a lot to learn. Can you honestly say that you how fix
something broken in systemd?

So I plan to migrate gradually and elogind is the way forward.
Gentoo maintenance for MATE is very slow but we will hopefully get there.
I hope someone in MATE want to take elogind on, it could be a way to drop Consolekit.

raveit65:
> Can you honestly say that you how fix something broken in systemd?

This is a very general question which i can't answer.
I am using systemd since the beginning with my fedora box.
I can't remember that there were problems which aren't fixed after a while.

> ..... it could be a way to drop Consolekit.

This sounds much better if it works with systemd.

vkareh:
I have no strong opinion on systemd, and I personally can't fix anything in systemd myself, but every distro I use supports it, which means someone will be assigned to work on it at some point. If mate-session-manager breaks because of something in systemd, I can fix that and test it to make sure it works. Unfortunately, elogind is not supported.

However, if elogind is a drop-in replacement for systemd, then all I need to worry about is having the right config flags and macros. As long as I don't need to modify the actual code (which I would not be able to test), then I'm fine with this.

I don't know much about the state of ConsoleKit, but if it's obsolete and unsupported, then it might be easier for us to have elogind as an alternative and drop ConsoleKit entirely. Or maybe we just leave it there until someone complains about it, then decide :P

sc0w:
If it works, my vote is to add elogind support, I think it can be supported by the community

Personally, I have systemd, but I am happy to see mate have a lot of users with multiple kind of systems.

lukefromdc:
I have no objection either, though I can only test on systemd as that's what I have installed on my bare-metal Debian Unstable machines

vkareh:
Excellent, thanks! :)

I'll leave this open for a few days to let anyone else who wants to test to do so. We can try to merge later this coming week.

joakim-tjernlund:
Later is here now :)
What do you think about mergin elogind now?

raveit65:
Can we add a comment? that it isn't official supported?
To avoid promising lies to users......

And there is a problem with Travis-CI

sc0w:
probably the travis error is because this PR have the **Draft** status ?

yetist:
@vkareh

Please do something to re-trigger travis, such as push, you may encounter travis bugs.

yetist:
This PR is #195, but travis think it is 194, maybe re-push can pass.

sc0w:
@yetist the PR 194 have the same name in title https://github.com/mate-desktop/mate-session-manager/pull/194 

where is handled the PR in mate-dev-scripts ?

sc0w:
... and both PRs have the same branch

yetist:
I think travis is wrong, if open travis from the Details link, it show the wrong PR number and commit id, and there is no build task of 195.

raveit65:
fixed, i will revert test file and do a push --force

raveit65:
Hmm, after i reverted the test commit with a `git push -f` the wrong build is back again.
I would simply close PR  and open a new one, to get rid of this madness.

raveit65:
Selecting the commit directly from travis side and triggering an build works fine.
So the problem is only the wrong PR number.
A new PR should not have the problem.

joakim-tjernlund:
Since you already tested Travis manually, why bother with a new PR and not just merge this one?

joakim-tjernlund:
@vkareh , just noticed that your modified PR moved AC_SUBST(HAVE_SYSTEMD) to 
before elogind. Not sure if that will work.

vkareh:
@joakim-tjernlund - this is what has been in testing for the last week, seems to work for everyone so far? my testings tell me that it works with `AC_SUBST` being called before updating the value, since it just exposes it to the rest of the code

joakim-tjernlund:
Ok, good.

vkareh:
the autoconf documentation is rather obscure, so I went by trial-and-error. Also, setting `AC_SUBST(HAVE_SYSTEMD)` as part of the ELOGIND checks seemed like a kludge.

vkareh:
Closed this in favor of #199 since GitHub's Draft PRs seem like they're breaking Travis-CI...

PR #200, opened 2019-03-22
Remove keyring from gnome-compat-startup key

### LINKED ISSUE #182: dconf setting /org/mate/desktop/session/gnome-compat-startup should not include 'keyring' by default
#### Expected behaviour

I wasted a whole day trying to get GnuPG's SSH agent working. Most of the things you read, e.g., [here](https://wiki.archlinux.org/index.php/GNOME/Keyring#Disable_keyring_daemon_components) tell you you need to prevent the SSH component of gnome-keyring from autostarting, which you can do by editing `/etc/xdg/autostart/gnome-keyring-ssh.service`, or doing the equivalent in `~/.config/autostart`. I would expect that to prevent the SSH component of gnome-keyring from being started on MATE.

#### Actual behaviour

Out of the box, none of this has any effect in MATE. It turns out this is because of a dconf property, `/org/mate/desktop/session/gnome-compat-startup`, which includes `"keyring"` by default. When this list contains `"keyring"`, `msm-gnome.c` is hardcoded to run `gnome-keyring-daemon --start`, which indiscriminately starts all components of `gnome-keyring-daemon` regardless of what you've done with the autostart stuff.

The reason I'm saying MATE shouldn't include it by default is because `gnome-keyring-daemon` includes the autostart files by default anyway, and they explicitly contain the line `OnlyShowIn=GNOME;Unity;`**`MATE;`**, which means that this "feature" is redundant. Maybe it could still be useful for distributions that don't include systemd? Even so, they're a minority, and they can write a patch to change the default dconf setting themselves if they want.

If you insist that this is the right default, I'll write a patch for NixOS at least.

Try to disable /etc/xdg/autostart's gnome-keyring-ssh.service, and observe that `gnome-keyring` continues to hijack the `SSH_AUTH_SOCK` environment variable no matter what you try to do.

#### MATE general version

1.21.0

#### Package version

`/nix/store/03f9d0cr3wpci47z9kx71a1386ffj3h5-mate-session-manager-1.21.0`

#### Linux Distribution

Right now, NixOS, but I had the same problem before on Debian, and unfortunately didn't remember how I solved it so I wasted a lot of time again today.

#### Link to downstream report of your Distribution

n/a

milouse:
Mate Desktop support /etc/xdg/autostart elements and gnome-keyring-daemon (GKD) add entries, which mention explicitely Mate desktop there. Thus this entry is redundant. On the other hand, this entry is responsible for a "start-all-subcomponent" of GKD, which block fine selection of components to load at startup.

I think this compat element is not needed any more.

Fix: https://github.com/mate-desktop/mate-session-manager/issues/182

### DIFF
--- a/data/org.mate.session.gschema.xml.in
+++ b/data/org.mate.session.gschema.xml.in
@@ -37,7 +37,7 @@
       <description>List of components that are required as part of the session. (Each element names a key under "/org/mate/desktop/session/required_components"). The Startup Applications preferences tool will not normally allow users to remove a required component from the session, and the session manager will automatically add the required components back to the session at login time if they do get removed.</description>
     </key>
     <key name="gnome-compat-startup" type="as">
-      <default>[ 'keyring', 'smproxy' ]</default>
+      <default>[ 'smproxy' ]</default>
       <summary>Control gnome compatibility component startup</summary>
       <description>Control which compatibility components to start.</description>
     </key>


raveit65:
Hmm, with this commit my ssh key isn't unlock anymore after session start.
```
[rave@mother caja-actions]$ git pull
Enter passphrase for key '/home/rave/.ssh/id_rsa': 
```
All three gnome-keyring in autostart are there and enabled.
```
[root@mother rave]# ls -l /etc/xdg/autostart/ | grep gnome-keyring
-rw-r--r--  1 root root  8355  2. Mär 19:15 gnome-keyring-pkcs11.desktop
-rw-r--r--  1 root root  7976  2. Mär 19:15 gnome-keyring-secrets.desktop
-rw-r--r--  1 root root  6538  2. Mär 19:15 gnome-keyring-ssh.desktop
```
```
[rave@mother caja-actions]$ gsettings get org.mate.session gnome-compat-startup
['smproxy']
```
Looks like this is needed.

milouse:
Hum… very weird… can you check if you see gnome-keyring-ssh.desktop in the « startup applications » gui ? If not, it may be disabled by something else. Do you have such an entry: `~/.config/autostart/gnome-keyring-ssh.desktop` ?

If not, it's surely a bug in gnome-keyring-daemon.

milouse:
Oh, no, forget about my previous message. Your behaviour is normal.

In fact, Mate Desktop will spawn `gnome-keyring-daemon --login --daemonize` when `keyring` is present in `gnome-compat-startup`. The `--login` switch implies the `--unlock` option by using the PAM password to unlocking compoment.

Without it, the later `gnome-keyring-daemon --start` will just enable component, without unlocking them.

Thus my PR is wrong. The problem seems not to be in mate itself, but because the `--login` switch somehow enable **all** components, instead of waiting for the corresponding `--start` call, as explained in the man page.

We should may be investigate on the GKD part for now.

milouse:
Mmmmh I just found something unclear to me. [This line](https://github.com/mate-desktop/mate-session-manager/blob/8f35c049172ab4bb1f73069e6a98f29ab45c4c0b/mate-session/msm-gnome.c#L99) seems to indicate that mate-desktop spawn `gnome-keyring-daemon --start` instead of `gnome-keyring-daemon --login --daemonize`.

I'm not a C export, nor a Mate guru :( @raveit65, if you have time, can you check if it works for you when you:

1. Keep the `gnome-compat-startup` value as `['keyring', 'smproxy']`
2. Replace the above line 99 by:
```C
argv[0] = GNOME_KEYRING_DAEMON;
argv[1] = "--login";
argv[2] = "--daemonize";
argv[3] = NULL;
```

Thank you very much!

raveit65:
> can you check if you see gnome-keyring-ssh.desktop in the « startup applications » gui ?

Yes, they are.

> Do you have such an entry: ~/.config/autostart/gnome-keyring-ssh.desktop ?

No
```
[rave@mother ~]$ ls -l .config/autostart/ | grep keyring
[rave@mother ~]$ 
```
With you last suggestion
```
--- a/mate-session/msm-gnome.c
+++ b/mate-session/msm-gnome.c
@@ -82,7 +82,7 @@ gnome_keyring_daemon_startup (void)
   gint         status;
   glong        pid;
   gchar       *end;
-  gchar       *argv[3];
+  gchar       *argv[4];
   gchar       *p;
   gchar       *name;
   const gchar *value;
@@ -96,8 +96,9 @@ gnome_keyring_daemon_startup (void)
 
   error = NULL;
   argv[0] = GNOME_KEYRING_DAEMON;
-  argv[1] = "--start";
-  argv[2] = NULL;
+  argv[1] = "--login";
+  argv[2] = "--daemonize";
+  argv[3] = NULL;
   g_spawn_sync (NULL, argv, NULL,
                 G_SPAWN_SEARCH_PATH | G_SPAWN_LEAVE_DESCRIPTORS_OPEN,
                 child_setup, NULL,
```
and
```
rave@mother ~]$ gsettings get org.mate.session gnome-compat-startup
['keyring', 'smproxy']
```
my login keyring won't be unlocked and after 10-15 sec a Warning pops up.
Screenshot at 2019-03-22 16-23-04
After i unlocked my keyring my ssh-key isn't unlock again.
```
[rave@mother caja-actions]$ git pull
Enter passphrase for key '/home/rave/.ssh/id_rsa': 
```
So it is much worse than before.

During testing my system crashes and i couldn't start firefox for some reasons.
Sorry , at the moment i am not amused.
Please test more for yourself, i am happy to test again a 100% tested PR.

milouse:
@raveit65 sorry for bothering you. In fact I'm not a mate developer and have no idea how to build and test part of mate desktop by myself. As you say in your first message that you were able to do so, I thought you were a Mate developer ready to help us.

I can try to test this by myself if you know where I can found some beginner information on how to build and test part of the desktop.

Thank you for your help and understanding.

auvipy (approved):

vkareh (changes requested):
This causes regressions. I think a possible alternative is for msm to parse the XDG autostart files to determine what should be started, and then spawn keyring with the `--start --components=[pkcs11,secrets,ssh]` flags based on those desktop files.

milouse:
This is a very old work of mine, and it’s a long time I do not use mate anymore. I prefer close this MR now and let anyone more involved work on it again if it’s still an issue (I assume it is not, given nobody never really look at this for all this time xD)

PR #202, opened 2019-03-27
Remove duplicate check for libelogind

vkareh:
Also default to no, since elogind is not officially supported.


vkareh:
@joakim-tjernlund - could you review/test this, please?

joakim-tjernlund:
Build tested --with-elogind, builds OK

PR #204, opened 2019-04-21
Add hardware-compatibility file and CFLAGS for mate_session_check_accelerated helpers

rbuj:


raveit65:
But the mentioned warning in other PR still remain.
```
** (mate-system-monitor:15597): WARNING **: 11:47:58.941: Unable to retrieve org.gnome.SessionManager.Renderer property.
```

PR #207, opened 2019-04-28
[Security] Use 'g_strlcpy' instead of 'strcpy'

yetist:
Potential insecure memory buffer bounds restriction in call 'strcpy', 
https://mate-session-manager.mate-desktop.dev/2019-04-28-021206-6482-1@b33479091537_clang-analyzer/report-d0d699.html#EndPath
https://mate-session-manager.mate-desktop.dev/2019-04-28-021206-6482-1@b33479091537_clang-analyzer/report-6da253.html#EndPath

### DIFF
--- a/mate-session/mdm.c
+++ b/mate-session/mdm.c
@@ -266,11 +266,11 @@ static gboolean mdm_init_protocol_connection(MdmProtocolData* data)
 
 	if (g_file_test(MDM_PROTOCOL_SOCKET_PATH, G_FILE_TEST_EXISTS))
 	{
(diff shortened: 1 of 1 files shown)


raveit65:
Does the effect only using mdm login-manger?
Because mdm was a linuxmint production only and recent linuxmint is switched to slick/lightdm login-manager.
Not sure if all this mdm code is really used nowadays.

raveit65:
@monsta 
Any idea if this code is used in general and/or by linuxmint?

raveit65:
As far i know MDM is only used by older linuxmint releases, which will never get a actual MATE update.
So this is
1. save to merge
2. code is obsolete, ihmo.

PR #212, opened 2019-05-02
main: Potential leak of memory pointed to by 'new_argv'

rbuj:


mate-session/main.c
```diff
@@ -524,9 +524,12 @@ static gboolean require_dbus_session(int argc, char** argv, GError** error)
 	if (!execvp("dbus-launch", new_argv))
 	{
 		g_set_error(error, G_SPAWN_ERROR, G_SPAWN_ERROR_FAILED, "No session bus and could not exec dbus-launch: %s", g_strerror(errno));
+		g_strfreev (new_argv);
 		return FALSE;
 	}
 
+	g_strfreev (new_argv);
```
muktupavels: This looks wrong... Looking at code this should be just `g_free`, no?

rbuj: Done

rbuj: Similar case at https://github.com/mate-desktop/mate-system-monitor/pull/158

PR #213, opened 2019-05-03
Hide NoDisplay Apps in Startup Applications

### LINKED ISSUE #134: Startup Applications doesn't respect NoDisplay=true
Several years ago, Ubuntu had an initiative to make the Startup Applications window empty by default. This was for two reasons:

1. To make it less intimidating
2. Many of these services should not be disabled by users. And if they are disabled, it results in bugs that may take some time to diagnose.

This was very easy to do: just add `NoDisplay=true` to the `/etc/xdg/autostart` files in the default install.

Since then, GNOME has gone one step further and gnome-session no longer builds the Startup Applications app, but Ubuntu patches it back in.

On Ubuntu (Unity) 16.04, the Startup Applications list should be empty. (Well, this has regressed a bit and there are actually 3 items but I've done some work to clean that up in 16.10 and 17.04 at least.)

But when I use Ubuntu MATE 16.04 and open System>Preferences>Personal>Startup Applications
there are a very long list of services. Many of them should not show up since `NoDisplay=true` is set.

Please do not show items in Startup Applications that have `NoDisplay=true`

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/41138931-startup-applications-doesn-t-respect-nodisplay-true?utm_campaign=plugin&utm_content=tracker%2F581048&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F581048&utm_medium=issues&utm_source=github).
</bountysource-plugin>

yetist:
Autostart applications are hidden in Startup Applications when NoDisplay=true.
Fix #134.

### DIFF
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -181,6 +181,10 @@ append_app (GsmPropertiesDialog *dialog,
             GspApp              *app)
 {
         GtkTreeIter   iter;
+        if (find_by_app (GTK_TREE_MODEL (dialog->list_store),
+                         &iter, app)) {
+                return;
+        }
 
         gtk_list_store_append (dialog->list_store, &iter);
         _fill_iter_from_app (dialog->list_store, &iter, app);
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -43,11 +43,13 @@
 #define CAPPLET_EDIT_WIDGET_NAME          "session_properties_edit_button"
 #define CAPPLET_SAVE_WIDGET_NAME          "session_properties_save_button"
 #define CAPPLET_REMEMBER_WIDGET_NAME      "session_properties_remember_toggle"
+#define CAPPLET_SHOW_HIDDEN_WIDGET_NAME   "session_properties_show_hidden_toggle"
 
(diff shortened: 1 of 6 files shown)


lukefromdc:
I see that mate-volume-contro-status-icon.desktop (which should be disabled if the panel applet is used instead for hidpi support is one of many that use NoDisplay=true . How are users supposed to disable default autostart apps without editing or removing .desktop files (which then is overwritten at every update) if they are hidden?

flexiondotorg:
I like this feature. I can understand why some might want to putting it behind a gsettings flag.

For the record, I will be hiding `NoDisplay=true` autostart desktop entries in Ubuntu MATE 🙂

yetist:
@lukefromdc @vkareh 

I added a checkbox and gsettings flag to control show or hide that apps.

flexiondotorg:
Shipping this in Ubuntu MATE 19.10. Distro patched and working a treat. Thank you 🙂

vkareh:
I think we could cherry-pick this to 1.22, no need for a distro patch

vkareh:
Done

PR #215, opened 2019-05-16
Gettext

yetist:


raveit65:
If you don't want to merge meson PRs, e.g. mozo,
it doesn't makes sense to merge/review gettext PRs for me, because `make dist` is broken with gettext.
And i need this for releasing developer releases.

yetist:
> If you don't want to merge meson PRs, e.g. mozo,
> it doesn't makes sense to merge/review gettext PRs for me, because `make dist` is broken with gettext.
> And i need this for releasing developer releases.

Mozo is different, it use new feature of meson-0.50.0, but the current version is 0.49.2 on debian/ubuntu, so it can build and create tarball on debian/ubuntu, but can create tarball on archlinux/fedora for release, it's enough. After user get the tarball, they can use autotools to install/uninstall.

The other repos may not need the new feature of meson, for example mate-backgrounds use meson-0.41.0.

lukefromdc:
We've already merged meson PR's for mate-desktop and mozo, so we need these now

raveit65:
You know that port to meson is needed after this?
With using 'gettext` the `make dist` command creates broken tarballs with .gmo files.
So we need `ninja -C build dist` to create tarballs.
See https://github.com/mate-desktop/mate-backgrounds/pull/22#issuecomment-492106967
for references.

raveit65:
Please check tarballs!
`make distcheck` creates .gmo and .po files in tarball.
That means we ship broken tarballs when this is merged.
The solution is to port to meson and using `ninja -C build dist` to create a tarball.
I suggest finishing this 2 ports for every repo before start porting to gettext in every repo,
and `make distcheck` is broken everywhere.

PS: marco is ported to gettext but meson port is missing.

lukefromdc:
I don't know enough about .gmo and .po files to make a valid check, unless we just want absence or presence of either one. I just checked to see that make-distcheck completed, then asked for other reviews. 

I agree with this and meson ports being one after the other given the issue with tarballs.

yetist:
Fix make dist to remove gmo files, result is here:
https://travis-ci.org/mate-desktop/mate-session-manager/jobs/549059011#L2214

raveit65:
Thanks again for this fix. Now we can merge this.

raveit65:
@yetist 
sorry, for asking you again, but we have 3 desktop files here which contain `icon` and `keyword` entry.
Can you add translators hints please?

PR #223, opened 2019-10-04
Fix timeout with gnome-keyring 3.34

jmakovicka:
Launch gnome-keyring-daemon asynchronously, and remove the
GNOME_KEYRING_PID reading code.

GNOME_KEYRING_PID exposure was removed from gnome-keyring in 2014.

### DIFF
--- a/mate-session/msm-gnome.c
+++ b/mate-session/msm-gnome.c
@@ -48,88 +48,46 @@
 
 
 static gboolean gnome_compat_started = FALSE;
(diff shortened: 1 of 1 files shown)


raveit65:
I can confirm that this fixes the problem in fedora-31-beta, but is this tested and save to use for older gnome versions?
I result we can use it for 1.22?

monsta:
Seems to work as before in Debian Stable as well (gnome-keyring is 3.20.0 there).

PR #230, opened 2020-02-18
main: allow turning on debugging from environment variable

romildo:
This commit lets users set the environment variable `MATE_SESSION_DEBUG` to 1 in order to enable debugging.

Having to pass the argument option `--debug` is pretty onerous for the user given the user doesn't normally run mate-session directly.

[gnome-session-manager](https://gitlab.gnome.org/GNOME/gnome-session/blob/master/gnome-session/main.c#L435) has a similar feature.

### DIFF
--- a/mate-session/main.c
+++ b/mate-session/main.c
@@ -606,6 +606,7 @@ int main(int argc, char** argv)
 	MdmSignalHandler* signal_handler;
 	static char** override_autostart_dirs = NULL;
 	gboolean gl_failed = FALSE;
+	const char *debug_string = NULL;
 
 	static GOptionEntry entries[] = {
 		{"autostart", 'a', 0, G_OPTION_ARG_STRING_ARRAY, &override_autostart_dirs, N_("Override standard autostart directories"), NULL},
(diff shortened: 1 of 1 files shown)


rbuj:
IMHO, It's not necessary to add a new environment variable, to enable debugging users can set G_DEBUG.
https://developer.gnome.org/glib/stable/glib-running.html

rbuj:
In addition, debugging can be enabled using dconf-editor:
https://github.com/mate-desktop/mate-session-manager/blob/bf37e0f2c55b78f6deaf3f9425ca164d84718037/mate-session/main.c#L657-L663

Screenshot at 2020-03-11 08-20-29

Environment variables / log level:
https://github.com/mate-desktop/mate-session-manager/blob/bf37e0f2c55b78f6deaf3f9425ca164d84718037/mate-session/main.c#L655

https://github.com/mate-desktop/mate-session-manager/blob/master/mate-session/mdm-log.c

rbuj:
And it can be also enabled using -debug option from command line:
https://github.com/mate-desktop/mate-session-manager/blob/3ab6fbfc811d00100d7a2959f8bbb157b536690d/mate-session/main.c#L613

raveit65:
Yes, it can be enabled via dconf, i hope it works......

raveit65:
No answer from author.

PR #255, opened 2020-06-28
Add compilation help

tamplan:
Use of submodule implies some commands before a compilation attempt.

### DIFF
--- a/README
+++ b/README
@@ -18,6 +18,11 @@ See https://wiki.gnome.org/Projects/SessionManagement/
 Installation
 ============
 
+In order to build this program from the source code you need git installed and
+ a working MATE environment, with the development tools installed properly.
+
+git submodule init
+git submodule update --remote --recursive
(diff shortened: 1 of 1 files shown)


README
```diff
@@ -18,6 +18,12 @@ See https://wiki.gnome.org/Projects/SessionManagement/
 Installation
 ============
 
+In order to build this program from the source code you need git installed and
+ a working MATE environment version 1.2, with the development tools installed
+properly.
```
raveit65, line 23: 1.2 ???
Where is this from? 
Damned, i over read that in mate-terminal PR.....

tamplan: Damned too, copy/paste error... I need to modify.

tamplan: Updated

raveit65: A tip, `git log -p` will show you the commit in terminal after commiting localy.

tamplan: Thanks, this could help.

PR #256, opened 2020-06-29
gsm-properties-dialog: Refactor notebook scroll event

rbuj:
Based on:
- https://github.com/mate-desktop/mate-terminal/commit/8d2eb0898154f2624e2ae34039169f307a3411a5
- https://github.com/mate-desktop/mate-terminal/commit/622e2d1272bf2d7a99d147482ae6a51b0a77f188

### DIFF
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -79,8 +79,6 @@ enum {
 };
 
 static void     gsm_properties_dialog_finalize    (GObject                  *object);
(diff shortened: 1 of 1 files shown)


raveit65:
I saw this issue in other repos too, and always restarting the build fixes the problem.
And this isn't a fedora only problem.
E.g https://travis-ci.org/github/mate-desktop/mate-session-manager/jobs/703148510
So, i don't think that a missing dependency is the real cause.

rbuj:
Ok, sorry for the inconvenience.

PR #257, opened 2020-06-29
Travis CI: fedora build: add xz

rbuj:


raveit65:
I saw this issue in other repos too, and always restarting the build fixes the problem.
And this isn't a fedora only problem.
E.g https://travis-ci.org/github/mate-desktop/mate-session-manager/jobs/703148510
So, i don't think that a missing dependency is the real cause.

rbuj:
Ok, sorry for the inconvenience.

sc0w:
@rbuj as core-team member, I think you have the rights to restart jobs at travis, you just need to login before

rbuj:
@sc0w yes, in fact I did it on other PR, but I don't recall on which ones I did it.

rbuj:
I can't restart the build because the "Restart build" button is missing:

Captura de Pantalla 2020-06-30 a les 12 10 40

raveit65:
Sometimes the restart buttons aren't visible for me too.
Causes are:
a) dns filtering by my vpn provider.
b) logout/login from travis helps a lot.

sc0w:
other issue with Travis, sometimes it doesn't show the green V in the list of commits, I don't know why

raveit65:
> other issue with Travis, sometimes it doesn't show the green V in the list of commits, I don't know why

Yes, maybe a problem with github or internet traffic between github and travis. Sometimes it won't be updated from yellow (pending) to green (success)

PR #261, opened 2020-11-03
Log Out / Shut Down dialogs: Restore classic GTK+ 2 appearance

thesquash:
While perusing [this thread on the Ubuntu MATE forum](https://ubuntu-mate.community/t/gtk3-regressions-from-a-gtk2-perspective/19511/56),
I noticed a group of many fellow MATE users who were discontented with how
GTK+ 3 styles message dialogs, the class of dialog used by the Log Out and
Shut Down dialogs in the MATE Session Manager.  The chief complaints made by
users were that GTK+ 3 does not display an icon next to the dialog contents,
that the buttons on the bottom of the dialog are overly large, and that --
possibly most importantly -- progress bars display their "progress text"
above the progress bar, not overlapping the progress bar as was the case in
GTK+ 2.  All manner of workarounds were proposed, and suffice it to say that
none were deemed good enough to attract considerable attention.

I tweaked the code for the Log Out and Shut Down dialogs considerably in this
commit to make the dialogs look very similar to the old GTK+ 2 dialogs of the
past.  First off, I subclassed GtkDialog (instead of GtkMessageDialog) and
added a grid to the dialog, placing an image and two label widgets inside the
grid to simulate the old GtkMessageDialog appearance.  Second, by means of
subclassing GtkDialog (as opposed to GtkMessageDialog), I gained the GTK+
2-esque button layout for free -- even in GTK+ 3 the button layout used by
GTK+ 2 dialogs is still used by default in GtkDialog but not GtkMessageDialog.
Third, I used a GtkProgressBar and GtkLabel packed into a GtkOverlay to
display the overlapping text, and a GtkSizeGroup to ensure the text is not
cut off by the small size of the progress bar.

Last but not least, GTK+ 3 progress bars have the frustrating property that
not the progress indicator, but instead a bunch of empty space, will appear
above them when they are allocated more screen space than their natural
size -- this is completely different from most GTK widgets which will expand
their "content" areas as the allocation is expanded, like the GtkButton
widget does.  This may not seem like a big deal, but it sure is if the user
set the text size larger than ~12pt (or with Adwaita, which has tiny progress
bars, probably .5pt).  In such a case the text will cover the entire progress
bar, and the top of the text will appear *above* the progress bar.  That does
not enhance accessibility!  To work around this, I used a kludge which
attaches a signal handler to the `size-allocate` signal emitted by the
GtkLabel.  When the signal handler is executed, it applies some simple CSS
to the progress bar, basically modifying the `min-height` property of the
progress indicator to equal the allocated height of the label:

        progressbar trough,
        progressbar progress
        {
          min-height: %dpx;
        }

The `%d` is replaced with the allocated height of the label.

Examples of the new look:

20201030-log-out
20201030-shut-down

And then as proof that the progress bars are resized properly, here's some extreme examples with 36-point font:

20201030-log-out-36pt-font
20201030-shut-down-36pt-font

### DIFF
--- a/mate-session/gsm-logout-dialog.c
+++ b/mate-session/gsm-logout-dialog.c
@@ -409,6 +409,7 @@ gsm_get_dialog (GsmDialogLogoutType type,
                 guint32             activate_time)
 {
         GsmLogoutDialog *logout_dialog;
+        GtkWidget       *button_box;
         GtkWidget       *grid;
         GtkWidget       *dialog_icon;
         GtkSizeGroup    *size_group;
@@ -425,6 +426,13 @@ gsm_get_dialog (GsmDialogLogoutType type,
 
         gtk_window_set_title (GTK_WINDOW (logout_dialog), "");
         gtk_window_set_resizable (GTK_WINDOW (logout_dialog), FALSE);
+
+        /*
+         * The following 2 lines are to stretch out the buttons at the bottom
+         * of the dialog, as is in vogue in GTK+ 3:
+         */
+        button_box = gtk_dialog_get_action_area (GTK_DIALOG (logout_dialog));
+        gtk_button_box_set_layout (GTK_BUTTON_BOX (button_box), GTK_BUTTONBOX_EXPAND);
 
         logout_dialog->type = type;
 
--- a/mate-session/gsm-logout-dialog.c
+++ b/mate-session/gsm-logout-dialog.c
@@ -2,7 +2,6 @@
  *
  * Copyright (C) 2006 Vincent Untz
  * Copyright (C) 2008 Red Hat, Inc.
- * Copyright (C) 2020 Gordon N. Squash.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License as
@@ -21,7 +20,6 @@
  *
  * Authors:
  *	Vincent Untz <vuntz@gnome.org>
- *      Gordon N. Squash
  */
 
 #include <config.h>
--- a/mate-session/gsm-logout-dialog.h
+++ b/mate-session/gsm-logout-dialog.h
@@ -1,7 +1,6 @@
 /* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*-
  *
  * Copyright (C) 2006 Vincent Untz
- * Copyright (C) 2020 Gordon N. Squash.
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License as
@@ -20,7 +19,6 @@
  *
  * Authors:
  *	Vincent Untz <vuntz@gnome.org>
- *      Gordon N. Squash
  */
 
 #ifndef __GSM_LOGOUT_DIALOG_H__
--- a/.build.yml
+++ b/.build.yml
@@ -67,7 +67,7 @@ requires:
     - make
     - mate-common
     - mesa-libGLES-devel
-    - pangox-compat-devel
+    - pango-devel
     - redhat-rpm-config
     - systemd-devel
     - xmlto
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,4 +71,4 @@ env:
   - DISTRO="archlinux:latest"
   - DISTRO="debian:testing"
   - DISTRO="fedora:latest"
-  - DISTRO="ubuntu:devel"
+  - DISTRO="ubuntu:rolling"


thesquash:
Weird.  According to Travis CI, this worked under 3 of the 4 test systems but failed under Fedora.  I suspect it's got something to do with an unrelated update to Fedora recently?

Pertinent part of the build log:

    >>> [docker exec -t mate-session-manager-fedora-build dnf install -y autoconf-archive clang clang-analyzer cppcheck-htmlreport dbus-glib-devel desktop-file-utils gcc git gtk3-devel libSM-devel libXtst-devel make mate-common mesa-libGLES-devel pangox-compat-devel redhat-rpm-config systemd-devel xmlto xorg-x11-xtrans-devel] <<<
    
    Last metadata expiration check: 0:00:33 ago on Tue Nov  3 21:48:53 2020.
    
    No match for argument: pangox-compat-devel
    
    Error: Unable to find a match: pangox-compat-devel

It looks like Fedora may have changed a package name or something.  I don't think my code is faulty -- it worked on every other test bed system.

With that said, any insight?  Did I actually slip up?

raveit65:
pangox-compat-devel is obsolete with fedora 33
It should build with using `pango-devel`
```
[root@mother rave]# rpm -qa pango-devel
pango-devel-1.44.7-2.fc32.x86_64
```
Can you commit please?
https://github.com/mate-desktop/mate-session-manager/blob/master/.build.yml#L70

raveit65:
Please rebase against master.
Build issue is fixed.
https://github.com/mate-desktop/mate-session-manager/commit/5460c692b1277c36847523e43b8ee8c19351ea22
https://github.com/mate-desktop/mate-session-manager/commit/1e97d140905dadb53d6839b32231ce2d2ccde85b

thesquash:
Sorry it took so long.  I was only half awake when I started the rebase!  Oh well, I hope I didn't *totally* mess something up...  :-(

Either way, thank you for the update to the Travis CI configuration.  As I write this the build test is in progress.

lukefromdc (approved):
I just tested this and it definately works. Here is how it looks with my UbuntuStudio_Legacy theme
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy

Logout_Dialog_test_11-4-2020

Although this look is different than any other progress bar and requires additional code, this dialog also has a function unique from any other: to shut down the session or the machine. I rather like this look myself. 

My opinion is that it reverses a change the  GTK3 transition rather than our intentional theming created so I endorse going with this, but the rest of the team also needs the opportunity to decide as this is a visual change.

Note that you have to start a new session to test this in action

raveit65 (dismissed):
Icon = OK
progressbar = OK
Button-style = No way!

I am missing the modern gtk3 button style. Please add the .linked style class to `GtkButtonBox`, which is the parent of the buttons
BlueMenta or Menta the default MATE gtk-theme use this style class for buttons in dialogs.

Buttons should be look like this:
Bildschirmfoto zu 2020-11-05 13-03-09

This looks wrong:
Bildschirmfoto zu 2020-11-05 13-05-19

Note: Gtk-themes which don't use `.linked button {}` are not affected 

Gtk-Inspector:
Bildschirmfoto zu 2020-11-05 13-23-11

mate-session/gsm-logout-dialog.c
```diff
@@ -20,6 +21,7 @@
  *
  * Authors:
  *	Vincent Untz <vuntz@gnome.org>
+ *      Gordon N. Squash
```
raveit65, line 24: Really?
I don't think that you are the author of this file.
One PR isn't enough to do that.
You're are not the maintainer of mate-session-manager!

mate-session/gsm-logout-dialog.h
```diff
@@ -1,6 +1,7 @@
 /* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*-
  *
  * Copyright (C) 2006 Vincent Untz
+ * Copyright (C) 2020 Gordon N. Squash.
```
raveit65, line 4: Same here.
You're are not the maintainer of mate-session-manger.

cwendling:
I can't test right now, but what about readability of the progress text with all themes?  IIUC the text is overlaid on the progress bar, which suggest it's not styled specifically to be readable there; for example what about a progress bar that is very dark (say, black) if the text is also black?

lukefromdc:
On additional testing I checked every theme I have installed, with these results:
All the MATE themes, Adwaita, Adwaita-dark, my own UbuntuStudio-Legacy theme, my old versions of Bluebird and Greybird, and the hight contrast themes all work with this. Bluementa, Blue-Submarine have somewhat reduced contrast but still plenty readable Only broken theme GTK's win32, which does not show the progress bar at all.

Still, I can see a potential issue here: if someone's theme were to set a very dark progress bar and black text the dialog would be unreadable. That said, I didn't find any such themes. Still, I suspect that may be why GTK devs moved text for this sort of widget to above rather than inside the progress bar.

As said before, other disadvantage is with this we get one "oddball" dialog. It would be unique among all our dialogs. Also I agree: this does NOT make the author a maintainer of mate-session-manager

thesquash:
I mostly agree with the proposed changes.  But there are a few things which I still think I should explain at least.

First off, as for the copyright issue, I understand I look like a jerk by inserting that.  I indeed agree to remove it.  But I was confused because Vincent Untz / Red Hat (the last copyright entries in the file) don't maintain these files either, and I believed at the time I had made a big enough change (subclassing a different widget) that I felt I should be at least held responsible for the changes.  That said, I looked my changes over again and I agree, so I removed the copyright notices and other disputable parts of the source files.

As for the dialog buttons:  On behalf of a large chunk of the Ubuntu MATE community, I respectfully disagree with the assertion that the "un-spread" dialog buttons are ugly.  I also disagree with the assertion that the dialog buttons look out of place -- MATE doesn't really have a uniform dialog button layout, quite frankly, but it appears the "un-spread" dialog buttons are in a slight minority, with traditional-looking dialogs having the older, "un-spread" layout:

**Atril:**

Atril_dialog

**Caja:**

Caja_delete_dialog
Caja_rename_dialog

**Engrampa:**

Engrampa_extraction_dialog
Engrampa_update_file_dialog

**Eye of MATE:**

EOM_dialog

**Marco** (ala Zenity -- note that this is, of all things, a GNOME dialog):

Marco_actually_Zenity_dialog

**MATE Panel:**

MATE-Panel-Force-Quit_dialog

**MATE System Monitor:**

MATE-System-Monitor-dialog

**Pluma:**

Pluma_dialog

(I didn't mean to overload anyone with too many screenshots.)  Many affiliated with the Ubuntu MATE project would go so far as to argue that the "spread" dialog buttons actually look unacceptable -- it's obviously a matter of opinion.  Furthermore, enabling the "spread" buttons actually uses a *deprecated* feature of GTK+ 3 (`gtk_dialog_get_action_area ()`) -- does anyone really want *more* warnings when compiling the code?

With that said, if you all truly despise the old style button layout, I've added another commit to expand the buttons, using the deprecated function of course.  However, many of us affiliated with Ubuntu MATE would greatly and strongly appreciate it if you would cherry-pick the commits and exclude the last one, if at all possible.

Thank you.

P.S. -- I tried adding the style class you suggested to the button box, but it still didn't look right at all.  This is not a theming thing; it's an implementation of the dialog thing.

lukefromdc:
I don 't use the .linked style class in 
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy
so not surprisingly I didn't see the initial linked/combined/spread button difference in my tests. I prefer separate buttons too: but with most of the MATE code, linking them is supported and it's up to the theme author to combine or separate them.

raveit65:
> P.S. -- I tried adding the style class you suggested to the button box, but it still didn't look right at all. This is not a theming thing; it's an implementation of the dialog thing.

Remove that commit please.
Here is an example to add a css style class with `gtk_style_context_add_class`
https://github.com/mate-desktop/mate-session-manager/blob/master/mate-session/gsm-logout-dialog.c#L144
```
        GtkStyleContext *context;
        context = gtk_widget_get_style_context (GTK_WIDGET (logout_dialog));
        gtk_style_context_add_class (context, "logout-dialog");
```
Do the same for the GtkButtonBox which is the parent of the buttons and name it `linked`.
That means themes like our default Menta or gtk default Adwaita use the `.linked button` style.
Other gtk themes like Mate-Ubuntu themes doesn't use that settings in their gtk-widget.css.

thesquash:
> Do the same for the GtkButtonBox which is the parent of the buttons and name it linked.

Even though that would use a deprecated function?  OK.

raveit65:
> > Do the same for the GtkButtonBox which is the parent of the buttons and name it linked.
> 
> Even though that would use a deprecated function? OK.

???
This function isn`t deprecated.
https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-add-class

mbkma:
TBH I think that the current dialogue looks nicer and cleaner. Not everything in GTK 2 was better :) Just my two cents...

> Last but not least, GTK+ 3 progress bars have the frustrating property that
> not the progress indicator, but instead a bunch of empty space, will appear
> above them when they are allocated more screen space than their natural
> size -- this is completely different from most GTK widgets which will expand
> their "content" areas as the allocation is expanded, like the GtkButton
> widget does. This may not seem like a big deal, but it sure is if the user
> set the text size larger than ~12pt (or with Adwaita, which has tiny progress
> bars, probably .5pt). In such a case the text will cover the entire progress
> bar, and the top of the text will appear above the progress bar. That does
> not enhance accessibility! To work around this, I used a kludge which
> attaches a signal handler to the size-allocate signal emitted by the
> GtkLabel. When the signal handler is executed, it applies some simple CSS
> to the progress bar, basically modifying the min-height property of the
> progress indicator to equal the allocated height of the label:

All of this is no problem in the current dialogue.

thesquash:
@raveit:  I was talking about `gtk_dialog_get_action_area ()` as the deprecated feature of GTK.

***

Well, I really don't care enough about this feature to add piles of code to restore a look that some of us like and some of us don't like.  As such I shall close this PR and carry on.  Soory to waste your time, everyone.

raveit65:
> I was talking about gtk_dialog_get_action_area () as the deprecated feature of GTK.

No idea why you think this is needed to add a simple css style class with `gtk_style_context_add_class ()`.

thesquash:
@raveit65:

If you show me the *exact code* you had in mind to add that style class to the button box without somehow using that deprecated function (`gtk_dialog_get_action_area ()`), *then* we can try this again.  But as it is, we're going around and around in circles misunderstanding each other over this simple situation.

TL;DR:  What's your idea for adding a "simple CSS style class" to the button box?  You're not thinking of adding it to the entire dialog instead of just the button box, are you?

raveit65:
I am stepping back from this review and i will patch this out for fedora when it gets merged.
Better ask ubuntu-mate to add it to their installer package.

lukefromdc:
Should we close this or keep it open?

thesquash:
Please close it. I've lost a lot of my energy for this cause. If I ever want to pursue it again I'll reopen it.

Thanks as always.

PR #265, opened 2021-01-18
Support xscreensaver when mate-screensaver is not available

rbuj:
Test: Suspend mate session

1. Click on System -> Shut Down...
2. Click on Suspend

### DIFF
--- a/mate-session/gsm-inhibit-dialog.c
+++ b/mate-session/gsm-inhibit-dialog.c
@@ -91,13 +91,24 @@ G_DEFINE_TYPE (GsmInhibitDialog, gsm_inhibit_dialog, GTK_TYPE_DIALOG)
 static void
 lock_screen (GsmInhibitDialog *dialog)
 {
-        GError *error;
-        error = NULL;
-        g_spawn_command_line_async ("mate-screensaver-command --lock", &error);
-        if (error != NULL) {
-                g_warning ("Couldn't lock screen: %s", error->message);
-                g_error_free (error);
+        gchar **screen_locker_command;
+
+        if ((screen_locker_command = gsm_get_screen_locker_command ()) != NULL) {
+                GError *error = NULL;
+
+                g_spawn_async (NULL, screen_locker_command, NULL, G_SPAWN_DEFAULT,
+                               NULL, NULL, NULL, &error);
+
+                if (error != NULL) {
+                        g_warning ("Couldn't lock screen: %s", error->message);
+                        g_error_free (error);
+                }
+
+        } else {
+                g_warning ("Couldn't find any screen locker");
         }
+
+        g_strfreev (screen_locker_command);
 }
 
 static void
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -1224,21 +1224,33 @@ sleep_lock_is_enabled (GsmManager *manager)
 static void
 manager_perhaps_lock (GsmManager *manager)
 {
-        GError   *error;
-        gboolean  ret;
+        gchar **screen_locker_command;
 
-        /* only lock if mate-screensaver is set to lock */
-        if (!sleep_lock_is_enabled (manager)) {
-                return;
+        if ((screen_locker_command = gsm_get_screen_locker_command ()) != NULL) {
+                GError *error = NULL;
+
+                /* only lock if mate-screensaver is set to lock */
+                if (!g_strcmp0 (screen_locker_command[0], "mate-screensaver-command") &&
+                    !sleep_lock_is_enabled (manager)) {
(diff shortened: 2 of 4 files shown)


lukefromdc (approved):
Had to start xscreensaver manually as it did not autostart with the session, but then it worked as expected. 


Locking screen from "lock screen" option in dialog involed when trying to suspend from the shut down button when power management inhibit (from the panel applet) is active WORKS

Locking screen from the main menu WORKS

I do not have my machine configured to lock the screen when coming out of suspend as this has been buggy in the past.

Note that we do not yet have an m-c-c entry to configure xscreensaver but probably could add one easily. The command to bring up the configuration gui for xscreensaver is xscreensaver-demo and the resulting gui does follow the GTK theme.

The unlocking dialog presented from the locked screen does not follow the GTK theme but this is intentional, to avoid linking against GTK and any crasher bugs that could come in with any past, present, or future version of GTK.

We could get a much simpler and more secure screensaver (less code linked to fewer libraries means much less to crash, and a screensaver crash unlocks the desktop) by porting the existing mate-screensaver themes to xscreensaver themes. Even OpenGL themes are supported. If the video driver crashes, the display freezes and an attacker cannot use the desktop of course. 

This PR handles integrating xscreensaver into the session, save that we need to start it automatically when mate-screensaver is not present. Fixing this, adding an m-c-c capplet to invoke xscreensaver-demo, and porting our screensaver themes to it would allow us to effectively offload screensaver maintainance to the xscreensaver team, yet have everything work the same save the appearance of the unlocking dialog-and fewer opportunities to unlock the whole machine by crashing the screensaver.

Approving as we may need to do this (or any desired portion of it) one step at a time. Best to fix xscreensaver startup now, but that too can be another PR if necessary

rbuj:
@lukefromdc You should only remove the last line of /etc/xdg/autostart/xscreensaver-autostart.desktop on Fedora:
```
OnlyShowIn=X-NODEFAULT;
```
https://src.fedoraproject.org/rpms/xscreensaver/blob/master/f/xscreensaver.spec#_955

Before
```
[robert@localhost ~]$ cat /etc/xdg/autostart/xscreensaver-autostart.desktop 
[Desktop Entry]
Type=Application
Name=xscreensaver-autostart
Comment=Autostart xscreensaver
Exec=/usr/libexec/xscreensaver-autostart
Terminal=false
X-Desktop-File-Install-Version=0.26
OnlyShowIn=X-NODEFAULT;
```
After
```
[robert@localhost ~]$ cat /etc/xdg/autostart/xscreensaver-autostart.desktop
[Desktop Entry]
Type=Application
Name=xscreensaver-autostart
Comment=Autostart xscreensaver
Exec=/usr/libexec/xscreensaver-autostart
Terminal=false
X-Desktop-File-Install-Version=0.26
```
/usr/libexec/xscreensaver-autostart content:
```
$ cat /usr/libexec/xscreensaver-autostart
#!/usr/bin/sh

# Don't launch xscreensaver if gnome-screensaver
# is installed

if [ -x /usr/bin/gnome-screensaver ] ; then exit 0 ; fi
if [ -x /usr/bin/mate-screensaver ] ; then exit 0 ; fi

exec xscreensaver -nosplash
```

rbuj:
https://bugzilla.redhat.com/show_bug.cgi?id=1918139

lukefromdc:
In Debian Unstable we don't have that file at all. It is of course possible to go m-c-c/Startup Applications and make a new item there to start xscreensaver with the session, about to test that. 

We could potentially ship a mate-xscreensaver-support package with an autostart file, an m-c-c entry for  xscreensaver-demo (to allow configuring screensaver options) and a port of our screensaver themes. This would give an almost identical user experience as now while offloading most screensaver maintainance and increasing screensaver security.

lukefromdc:
Does anyone else what to review this? If not, should I merge it or ignore it?

rbuj:
This pull request is still required since it fixes the error described in the title. It's required for all operating systems, but especially for onpenindiana which only includes xscreensaver.

The bug I opened in Red Hat is for fixing another bug that prevents Fedora users from testing this bugfix, as by default the automatic boot of the xscreensaver daemon is disabled for almost all desktop environments from Fedora 27+.

lukefromdc:
OK, are we OK to merge this or should we wait for another reviewer?

rbuj:
I think it's ready to go. 

This weekend I will start posting the pull requests for removing mate-screensaver support on all projects. Since we will probably have to close mate-screensaver for copyright issues, as you can see in the blog post comments https://www.jwz.org/blog/2021/01/i-told-you-so-2021-edition/

These last few days I've been checking the mate-settings-daemon code.
I will postpone the revision of the managers, which are no thread safe singletons. (new todo task)

PR #266, opened 2021-01-23
mate-session: remove mate-screensaver support

rbuj:
Related mate-desktop/mate-screensaver#243


lukefromdc:
I assume we can close this?

Issue #156, opened 2018-03-29
Services in autostart can't be disabled


ceccoemi:
I'm not sure if it is an issue related to mate-session or to mate-screensaver, please let me know if I'm wrong.

I noticed that if I remove the program mate-screensaver from the list of startup processes (from the GUI program or manually changing the file in `/etc/xdg/autostart/mate-screensaver.desktop`) it will continue to be launched at login. It will happen even if I completely remove the files `/etc/xdg/autostart/mate-screensaver.desktop` and `~/.config/autostart/mate-screensaver.desktop`. I noticed that it will not be launched if I remove the file `/usr/share/dbus-1/services/org.mate.ScreenSaver.service`, but probably that solution is not applicable

#### How to reproduce

- Open the program Startup Applications
- Untick the section with Screensaver
- Log out
- Log in
- Open a terminal and run `grep -f mate-screensaver`, it will be displayed the pid.

#### Distro and Desktop version

I have Ubuntu MATE 17.10 with MATE Desktop 1.18.0, but I found this bug even in other MATE versions in Ubuntu 16.04 (in real hardware) and Ubuntu 18.04 (in a virtualbox)

L-U-T-i:
I am noticing the same issue with the latest git commit builds (master / 1.25). Had to uninstall mate-screensaver to get rid of mate-screensaver processes...

raveit65:
Not only mate-screensaver is affected by this issue. Every application in autostart can't be disabled.
Not sure, but IHMO we have introduce this issue with one of those PRs.
https://github.com/mate-desktop/mate-session-manager/pull/184
https://github.com/mate-desktop/mate-session-manager/pull/213
https://github.com/mate-desktop/mate-session-manager/pull/178

But there is a workaround.
After un-checking the checkbox, open the edit dialog of the autostart service and click on the save button in dialog.
This should save the new state of the service.
Try it out.

So this is definitely a bug in mate-session-manager.

L-U-T-i:
> Not only mate-screensaver is affected by this issue. Every application in autostart can't be disabled.
> Not sure, but IHMO we have introduce this issue with one of those PRs.
> #184
> #213
> #178
> 

It seems to be [#213](https://github.com/mate-desktop/mate-session-manager/pull/213) - if I revert that PR, I can enable and disable screensaver (my test example) without a workaround below. Furthermore, if I revert just a part of that PR (only _gsm-properties-dialog.c_ changes):
[...]

PR #272, opened 2021-04-03
autostart: fix disabling services in dialog

raveit65:
- fixes https://github.com/mate-desktop/mate-session-manager/issues/156

Steps to reproduce the issue and testing the fix see https://github.com/mate-desktop/mate-session-manager/issues/156#issuecomment-808731386
and
https://github.com/mate-desktop/mate-session-manager/issues/156#issuecomment-808893461

### DIFF
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -179,10 +179,6 @@ append_app (GsmPropertiesDialog *dialog,
             GspApp              *app)
 {
         GtkTreeIter   iter;
-        if (find_by_app (GTK_TREE_MODEL (dialog->list_store),
-                         &iter, app)) {
-                return;
-        }
 
         gtk_list_store_append (dialog->list_store, &iter);
         _fill_iter_from_app (dialog->list_store, &iter, app);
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
I just tested this, and had no trouble turning either my custom startup jobs or mate-session-manager on and off

lukefromdc:
This one's important, we need to get this finished and merged before the unstable release

raveit65:
Yes, this also needs to be go in 1.24 branch.

rbuj:
IMHO, I don't think it's a good idea to remove the check to avoid adding the app if it already exists from `append_app` method. I think we need to review the `find_by_app` method where it checks that the apps are not the same, as it currently compares the pointers of the objects instead of their content.

raveit65:
I only took that code from https://github.com/mate-desktop/mate-session-manager/issues/156#issuecomment-809557273
But feel free to find a better solution to fix that issue.
Btw. i used that patch already in latest fedora package, in result the issue isn't reproducible with standard installation.

rbuj (changes requested):
It adds a regression as it shows all applications after launching `mate-session-properties` because it removes `update_tree_view` call after calling `populate_model`.

raveit65:
@rbuj 
again, I only took that code from #156 (comment)
But feel free to find a better solution to fix that issue when you know it better.

PR #274, opened 2021-04-17
capplet: use a function to set the visibility of treeview rows

### LINKED ISSUE #260: XDG autostart enable/disable functionality not compatible with XFCE
#### Expected behaviour

.desktop files in XDG autostart locations with the key Hidden= set to true are visible in the mate-session-properties dialog and can be (re-)enabled there, other desktop envs use Hidden=true|false to enable/disable startup applications. See also #134 for how to hide startup applications from the mate-session-properties dialog entirely.

#### Actual behaviour

Currently, mate-session-properties uses a MATE specific flag for disabling XDG autostart applications (X-MATE-Autostart-enabled=false). This boolean switch is not portable to other desktop environments. Please consider using the Hidden= key instead.

The explicit problem I have is this:

  * disable autostart of an application in XFCE -> this sets Hidden=true in the .desktop file
  * I cannot re-enable autostart for this application in MATE

Other way around:

  * I disable autostart of an application in MATE -> this sets X-MATE-Autostart-enabled=false
  * the application still autostarts in XFCE

I haven't tested other desktop envs. Let me know, if I should.

The context is: Ayatana Indicators. The system indicators are services usable across different desktop envs. At some point, they should be controllable via systemd user services (on systems having systemd). On systems without systemd / as of today, XDG autostart is used to launch the system indicator services. It should be possible to enable/disable those services cross-desktop-env.

Install MATE and XFCE on the same machine. Disable/enable XDG autostarted application using mate-session-properties and XFCE's xfce4-session-settings and switch between desktop environments

#### MATE general version

1.20 / 1.24

#### Package version

1.20

#### Linux Distribution

Debian/Ubuntu

#### Link to downstream report of your Distribution

none

rbuj:
closes #260

### DIFF
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -116,7 +116,7 @@ _fill_iter_from_app (GtkListStore *list_store,
         const char *description;
         const char *app_name;
 
-        enabled     = gsp_app_get_enabled (app);
+        enabled     = !gsp_app_get_hidden (app);
         icon        = gsp_app_get_icon (app);
         description = gsp_app_get_description (app);
         app_name    = gsp_app_get_name (app);
@@ -268,10 +268,9 @@ on_startup_enabled_toggled (GtkCellRendererToggle *cell_renderer,
                             -1);
 
         active = gtk_cell_renderer_toggle_get_active (cell_renderer);
-        active = !active;
 
         if (app) {
-                gsp_app_set_enabled (app, active);
+                gsp_app_set_hidden (app, active);
                 g_object_unref (app);
         }
 }
@@ -556,12 +555,10 @@ visible_func (GtkTreeModel *model,
                             -1);
 
         if (app) {
-                gboolean hidden;
                 gboolean nodisplay;
 
-                hidden = gsp_app_get_hidden (app);
                 nodisplay = gsp_app_get_nodisplay (app);
-                visible = !hidden && (show_hidden || !nodisplay);
+                visible = show_hidden || !nodisplay;
                 g_object_unref (app);
         } else {
                 visible = show_hidden;
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -69,8 +69,7 @@ struct _GsmPropertiesDialog
 };
 
 enum {
-        STORE_COL_VISIBLE = 0,
-        STORE_COL_ENABLED,
+        STORE_COL_ENABLED = 0,
         STORE_COL_GICON,
         STORE_COL_DESCRIPTION,
         STORE_COL_APP,
@@ -112,13 +111,11 @@ _fill_iter_from_app (GtkListStore *list_store,
                      GtkTreeIter  *iter,
                      GspApp       *app)
 {
-        gboolean    hidden;
         gboolean    enabled;
         GIcon      *icon;
         const char *description;
         const char *app_name;
 
-        hidden      = gsp_app_get_hidden (app);
         enabled     = gsp_app_get_enabled (app);
         icon        = gsp_app_get_icon (app);
         description = gsp_app_get_description (app);
@@ -150,7 +147,6 @@ _fill_iter_from_app (GtkListStore *list_store,
         }
 
         gtk_list_store_set (list_store, iter,
-                            STORE_COL_VISIBLE, !hidden,
                             STORE_COL_ENABLED, enabled,
                             STORE_COL_GICON, icon,
                             STORE_COL_DESCRIPTION, description,
@@ -196,13 +192,13 @@ _app_added (GsmPropertiesDialog *dialog,
             GspApp              *app,
             GspAppManager       *manager)
 {
+        (void) manager;
         append_app (dialog, app);
 }
 
 static void
-_app_removed (GsmPropertiesDialog *dialog,
-              GspApp              *app,
-              GspAppManager       *manager)
+remove_app (GsmPropertiesDialog *dialog,
+            GspApp              *app)
 {
         GtkTreeIter iter;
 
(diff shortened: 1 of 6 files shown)


raveit65:
For what is this ?

rbuj:
It improves the code comprehension and it removes unused arguments.

The `_app_added` and `_app_removed` methods are used for updating the applications listed on the treeview when there is a change in the manager. The `update_tree_view` method adds or removes the applications which are not listed by default on the treeview (see https://github.com/mate-desktop/mate-session-manager/commit/16fe21a84ea333db67e9f0509d619f8d3fee1443)

raveit65:
Does it fix this report?
https://github.com/mate-desktop/mate-session-manager/issues/156

rbuj:
No it isn't. It's a frontend bugfix.

#156 may be related to the backend, ie manager. I haven't reviewed it yet:
- adding or removing items in the view could call the manager
- the manager has not been implemented to handle this kind of apps (no display: introduced at 16fe21a)
- ...

capplet/gsm-properties-dialog.c
```diff
@@ -217,6 +216,14 @@ _app_removed (GsmPropertiesDialog *dialog,
         gtk_list_store_remove (dialog->list_store, &iter);
 }
 
+static void
+_app_removed (GsmPropertiesDialog *dialog,
+              GspApp              *app,
+              GspAppManager       *manager)
```
raveit65, line 222:
```
gsm-properties-dialog.c:222:36: warning: unused parameter 'manager' [-Wunused-parameter]

              GspAppManager       *manager)
```
I know `_app_removed` was there before, but now you write  the code snipset new.

rbuj: Ok, I updated the pr because i think it's easier to use a function than a column for setting the visibility of the treeview rows.

raveit65:
I tested all 3 PRs from you and my PR before you updated this PR, and the result was that the fix for https://github.com/mate-desktop/mate-session-manager/issues/156 doesn't work proper any more.
With current changes we have a conflict with my PR and one of yours.

For me the important thing is to have a proper fix for https://github.com/mate-desktop/mate-session-manager/issues/156 which was done with https://github.com/mate-desktop/mate-session-manager/pull/272

rbuj:
#272 adds a regression as it shows all applications after launching the application because it removes update_tree_view call after calling populate_model.

raveit65:
Mate-screensaver is always shown in dialog since the begining of MATE desktop.
This isn't wrong!

rbuj:
@raveit65 I mean the "show hidden" checkbox.

raveit65:
When i only use https://github.com/mate-desktop/mate-session-manager/pull/272
i can disable hidden application and i can show hidden applictions in my test.
And with both values i can disable/enable a start of an application.
I don't get what is wrong in real testing.
Don't understand me wrong....
If you really did test my PR and think this is wrong than please provide a better solution to fix https://github.com/mate-desktop/mate-session-manager/issues/156.
I am not against an improvement.

But at the moment i see only 3 new PRs but enable/disable applications in autostart isn't fixed.

rbuj:
Try to disable `Bluetooth OBEX Agent`, which is not shown if the checkbox "show hidden" is not checked:
```
$ grep NoDisplay /etc/xdg/autostart/blueberry-obex-agent.desktop
NoDisplay=true
```
After disabling it, it should create the file below:
```
$ diff .config/autostart/blueberry-obex-agent.desktop /etc/xdg/autostart/blueberry-obex-agent.desktop
13d12
< X-MATE-Autostart-enabled=false
```
Log out/in, check blueberry-obex-agent:
```
$ grep Exec /etc/xdg/autostart/blueberry-obex-agent.desktop
Exec=/usr/lib/blueberry/blueberry-obex-agent.py
$ ps ax | grep blueberry-obex-agent
```

rbuj:
Ok, I think the bug you're referring to is that it doesn't save the change after you clicked the the checkbox, and it's only saved if when you clicked on another item of list.

rbuj:
`on_startup_enabled_toggled` is working properly.

raveit65:
The bug is that enable/disable a service in main GUI doesn't change the settings, a workaround is to open the preferences of `a service` and save the status of a service there.
The issue only exists if `enable-hidden` checkbox is checked.
Described here:
https://github.com/mate-desktop/mate-session-manager/issues/156#issuecomment-808731386
https://github.com/mate-desktop/mate-session-manager/issues/156#issuecomment-808893461

raveit65:
> and it's only saved when you clicked on another item of list.

Looks like you found another workaround :)

rbuj:
> > and it's only saved when you clicked on another item of list.
> 
> Looks like you found another workaround :)

It's because of GSP_APP_SAVE_DELAY. There is a delay of two second before saving the file.

Anyway the files are generated in `$HOME/.config/autostart` folder when there is no difference with the original file, and they are deleted otherwise.

capplet/gsm-properties-dialog.c
raveit65: The warning is still there.
https://travis-ci.org/github/mate-desktop/mate-session-manager/jobs/767534724#L2537
```
gsm-properties-dialog.c:218:36: warning: unused parameter 'manager' [-Wunused-parameter]

              GspAppManager       *manager)
```

raveit65:
I can confirm that this PR fixes https://github.com/mate-desktop/mate-session-manager/issues/156 and other open variants of the same issue. Thank you
Which part fixes enable/start services?
We need to fix the issue for 1.24 branch too.

capplet/gsm-properties-dialog.c
rbuj: done

raveit65: Thx

raveit65 (approved):
LGTM

raveit65:
@rbuj 
Simply cherry-picking doesn't work for stable branch.
Can this be done for 1.24 branch, please?

rbuj:
ok, wip

rbuj:
PR #279

PR #275, opened 2021-04-18
capplet: redundant cast GTK_TREE_MODEL, unnecessary local variables, …

rbuj:
…unused null assignment

### DIFF
--- a/capplet/gsm-properties-dialog.c
+++ b/capplet/gsm-properties-dialog.c
@@ -251,14 +251,11 @@ on_startup_enabled_toggled (GtkCellRendererToggle *cell_renderer,
         GspApp     *app;
         gboolean    active;
 
-        if (!gtk_tree_model_get_iter_from_string (GTK_TREE_MODEL (dialog->tree_filter),
-                                                  &iter, path)) {
+        if (!gtk_tree_model_get_iter_from_string (dialog->tree_filter, &iter, path)) {
                 return;
         }
 
-        app = NULL;
-        gtk_tree_model_get (GTK_TREE_MODEL (dialog->tree_filter),
-                            &iter,
+        gtk_tree_model_get (dialog->tree_filter, &iter,
                             STORE_COL_APP, &app,
                             -1);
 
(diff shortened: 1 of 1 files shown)


capplet/gsm-properties-dialog.c
```diff
@@ -583,28 +572,25 @@ setup_dialog (GsmPropertiesDialog *dialog)
...
                                                  G_TYPE_STRING);
-        tree_filter = gtk_tree_model_filter_new (GTK_TREE_MODEL (dialog->list_store),
-                                                 NULL);
-        g_object_unref (dialog->list_store);
-        dialog->tree_filter = tree_filter;
 
-        gtk_tree_model_filter_set_visible_column (GTK_TREE_MODEL_FILTER (tree_filter),
+        dialog->tree_filter = gtk_tree_model_filter_new (GTK_TREE_MODEL (dialog->list_store),
+                                                         NULL);
+
```
raveit65, line 578: Is this missing?
```
g_object_unref (dialog->list_store);
```

rbuj: No it isn't. It is not unreferenced in the code below:

https://gitlab.gnome.org/GNOME/gtk/-/blob/master/testsuite/gtk/filtermodel.c#L125-154
```C
static void
filter_test_setup_generic (FilterTest    *fixture,
                           gconstpointer  test_data,
                           int            depth,
                           gboolean	  empty,
                           gboolean       unfiltered)
{
  const GtkTreePath *vroot = test_data;
  GtkTreeModel *filter;

  fixture->store = create_tree_store (depth, !empty);

  g_signal_connect_swapped (fixture->store, "row-changed",
                            G_CALLBACK (filter_test_store_signal), fixture);

  /* Please forgive me for casting const away. */
  filter = gtk_tree_model_filter_new (GTK_TREE_MODEL (fixture->store),
                                      (GtkTreePath *)vroot);
  fixture->filter = GTK_TREE_MODEL_FILTER (filter);

  if (!unfiltered)
    gtk_tree_model_filter_set_visible_column (fixture->filter, 1);

  /* We need a tree view that's listening to get ref counting from that
   * side.
   */
  fixture->tree_view = gtk_tree_view_new_with_model (filter);

  fixture->monitor = signal_monitor_new (filter);
}
```

rbuj:
The ui file can be promoted to a dialog and we could use a template for binding the widgets and their event handlers.

PR #276, opened 2021-04-18
capplet: memory leak

rbuj:


capplet/gsm-properties-dialog.c
```diff
@@ -102,6 +102,8 @@ find_by_app (GtkTreeModel *model,
                         g_object_unref (iter_app);
                         return TRUE;
                 }
+
+                g_object_unref (iter_app);
```
raveit65, line 106: Isn't it already unref in line102?
But maybe i don't understand the code well....

rbuj: Currently, it's only unreferenced when both objects are equal before returning TRUE.
We should unreference all returned objects by gtk_tree_model_get.

> Returned values with type G_TYPE_OBJECT have to be unreferenced, values with type G_TYPE_STRING or G_TYPE_BOXED have to be freed. Other values are passed by value.

https://developer.gnome.org/gtk3/stable/GtkTreeModel.html#gtk-tree-model-get

raveit65: OK

raveit65 (approved):
LGTM

mbkma (approved):
From devhelp on `gtk_tree_model_get`:

`Returned values with type G_TYPE_OBJECT have to be unreferenced, values with type G_TYPE_STRING or G_TYPE_BOXED have to be freed. Other values are passed by value.`

And `GspApp` is actually of type `G_TYPE_OBJECT`:

`capplet/gsp-app.c:G_DEFINE_TYPE_WITH_PRIVATE (GspApp, gsp_app, G_TYPE_OBJECT)`

PR #278, opened 2021-04-21
Look for the location of the locking screen app using user's PATH

### LINKED ISSUE #277: screen not locked after suspend
#### Expected behaviour
I'm on Fedora 33. And until the update to mate-session-manager 1.24.2 the screen was locked properly after resuming from suspend. My mate-screensaver is set to lock on blank.

#### Actual behaviour
Screen is not locked and the journal contains:
mate-session[9518]: WARNING: Couldn't lock screen: Failed to execute child process “mate-screensaver-command” (No such file or directory)

I assume this happens on all Fedora 33 and other distributions as well after updating mate-session-manager containing
https://github.com/mate-desktop/mate-session-manager/commit/071933646ef80fff6295b581376d018ba8ddc39a

#### MATE general version
1.24.1
mate-desktop-1.24.1-3.fc33.x86_64

#### Package version
mate-session-manager-1.24.2-2.fc33.x86_64

#### Linux Distribution
Fedora 33

#### Link to bugreport of your Distribution (requirement)
I report this only here because it looks like something got wrong in
https://github.com/mate-desktop/mate-session-manager/commit/071933646ef80fff6295b581376d018ba8ddc39a
and mate-screensaver-command is not found while it is installed:
```
# which mate-screensaver-command 
/usr/bin/mate-screensaver-command
```
and working as expected when called with "--lock".

rbuj:
closes #277

### DIFF
--- a/mate-session/gsm-inhibit-dialog.c
+++ b/mate-session/gsm-inhibit-dialog.c
@@ -96,7 +96,8 @@ lock_screen (GsmInhibitDialog *dialog)
         if ((screen_locker_command = gsm_get_screen_locker_command ()) != NULL) {
                 GError *error = NULL;
 
-                g_spawn_async (NULL, screen_locker_command, NULL, G_SPAWN_DEFAULT,
+                g_spawn_async (NULL, screen_locker_command, NULL,
+                               G_SPAWN_DEFAULT | G_SPAWN_SEARCH_PATH,
                                NULL, NULL, NULL, &error);
 
                 if (error != NULL) {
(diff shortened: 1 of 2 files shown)


raveit65:
Thanks for quick reaction.

raveit65:
@MASHtm
Does this works for you?

MASHtm:
Yes, I added your patch to the f33 spec file and now it works as expected. Thanks for the quick fix.

raveit65 (approved):
LGTM

PR #279, opened 2021-04-22
capplet: use a function to set the visibility of treeview rows (1.24 branch)

rbuj:


raveit65 (approved):
Works fine with 1.24 branch.
Thanks

PR #282, opened 2021-05-17
util: Blacklist some session-specific variables

### LINKED ISSUE #271: Shouldn't add XDG_SESSION_ID to dbus and systemd activation environment
* Install a Debian 11 system with gdm 3.38, GNOME 3.38, MATE 1.24.x and systemd
* Log in via ssh (for example assume systemd-logind assigns login session ID 2 for this)
* Log in to MATE, as the same user (for example assume systemd-logind assigns login session ID 4 for this)
* Log out from MATE
* Log in to GNOME, as the same user, without rebooting (for example assume systemd-logind assigns login session ID 6 for this)
* Lock the screen
* Try to unlock the screen

#### Expected behaviour
* When I log in to MATE, `mate-session-manager` uploads some environment variables to `dbus-daemon --session` via `UpdateActivationEnvironment`, and either `mate-session-manager` itself or `dbus-daemon` passes on those environment variables to `systemd --user` via `SetEnvironment`
* Because the `systemd --user` and `dbus-daemon --session` processes are shared between multiple login sessions, these environment variables **should not** include things that are highly specific to one login session, and in particular `XDG_SEAT`, `XDG_SESSION_ID` and `XDG_VTNR`
* Processes run by `systemd --user`, including most of GNOME, do not inherit `XDG_SESSION_ID`
* In particular, GNOME Shell does not inherit `XDG_SESSION_ID`, correctly determines its own systemd-logind session ID (in my example, session ID 6) by querying systemd-logind, and can lock and unlock correctly

#### Actual behaviour
* `mate-session-manager` sends **all** environment variables to `systemd --user`, including `XDG_SESSION_ID=4` in my example
* Because the ssh login session keeps `systemd --user` alive, `XDG_SESSION_ID=4` stays in the activation environment and is "leaked" into the subsequent GNOME session
* GNOME Shell inherits `XDG_SESSION_ID=4` from `systemd --user`, even though that login session has already ended
* This makes GNOME Shell try to communicate with a logind session that no longer exists, breaking the ability to unlock the screen
    * I contributed a change to `gnome-session` to work around this, but I don't think it's really a GNOME bug

#### MATE general version
1.24.1 (not tested with 1.24.2, but none of the changes since 1.24.1 seem like they would affect this)

#### Package version
1.24.1-1

#### Linux Distribution
Debian 11

#### Link to bugreport of your Distribution (requirement)
* MATE bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986100

---

https://github.com/linuxmint/cinnamon-session/pull/141#issuecomment-809590165 lists some other related fixes in the ancestor project gnome-session. It would be a good idea to incorporate those into mate-session-manager at the same time.

Related bug reports:
* equivalent Cinnamon bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=986099, https://github.com/linuxmint/cinnamon-session/issues/140, https://github.com/linuxmint/cinnamon-session/pull/141
* original bug report against GNOME: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=973474
* workaround in GNOME: https://gitlab.gnome.org/GNOME/gnome-session/-/issues/86

sunweaver:
Things like XDG_SESSION_ID should not be uploaded to the environment.
For example this is broken currently:

  1. SSH to your machine
  2. Log in to MATE Shell
  3. Log out
  4. Log in again
  5. Lock the screen
  6. Try to unlock

You can't, and this is because the XDG_SESSION_ID from the first session
(step 2) has leaked through to the second one (step 4), and so MATE
Shell is listening to the `logind` `UnlockSession` signal for the wrong
session. The SSH session established in step 1 serves to keep the
`systemd --user` instance alive, so that the state is not torn down
between logins.

Original patch for GNOME by Iain Lane <iainl@gnome.org>.

Patch ported over to MATE's session manager by Mike Gabriel
<sunweaver@debian.org>.

Fixes https://github.com/mate-desktop/mate-session-manager/issues/271


sunweaver:
@raveit65 This PR has a failing CI build. However, this does not look like a PR / patch problem.

vkareh (approved):

PR #284, opened 2021-10-21
Fix memory leaks

rbuj:


lukefromdc:
How are these tools best tested?

lukefromdc:
No build warnings from these files, and the code change is very simple, four lines freeing variables when we are done with them. If the new session I am about to start up (w compiz and the AMD driver) doesn't segfault than this should be good to go.,

lukefromdc (approved):
New session with these installed works fine, would expect a segfault if we got a use after free from this. None showing up in dmesg

thesquash (approved):
These changes are fine, and they work for me too (on Intel integrated graphics hardware).  That's to be expected since none of the things that now get freed in this PR get used again; they are all freed right before each stub program exits.  As such there isn't any real memory leak problem, but hey, I guess it makes the code cleaner and more proper, right?

raveit65:
Debian Build at travis-Ci in master failed after merging this PR.
https://app.travis-ci.com/github/mate-desktop/mate-session-manager/builds/254626655

lukefromdc:
Here again at Travis we have this:

```
 GEN      gsm-presence-glue.h

Compilation failed: Failed to close file descriptor for child process (Operation not permitted)

make[2]: *** [Makefile:1313: gsm-manager-glue.h] Error 1

make[2]: *** Waiting for unfinished jobs....

Compilation failed: Failed to close file descriptor for child process (Operation not permitted)
```
as would be expected if the docker image is the problem

PR #288, opened 2021-11-23
g_key_file_save_to_file () is available from glib 2.40

rbuj:
https://bugzilla.gnome.org/show_bug.cgi?id=309224


raveit65 (approved):
LGTM

lukefromdc (commented):
This builds fine on Debian Unstable, but I have no idea how to test it as I don't have anything going on using keyfiles that I know of (unless this affects the login keyring when only a passphrase is used to log back in after logout)

zhuyaliang (approved):
Agree to use ```g_key_file_save_to_file ```

PR #289, opened 2021-11-23
gsm-manager: use g_settings_schema_source_lookup ()

rbuj:


lukefromdc (approved):
Builds fine, no issues starting a new session with this installed, and mate-screensaver works as normal with this. Not sure how else to test this

raveit65 (approved):
A better screensaver detection,
builds and runs fine.

PR #299, opened 2022-07-25
configure.ac: The macro 'AC_HELP_STRING' is obsolete

rbuj:


lukefromdc (approved):
configuration and running ./configure --help work just fine with this.

AS_HELP_STRING has been around back to 2008 at LEAST, as that's when Autoconf 2.62 came out. Thus no distro should get a build failure from this change.

Anyone know of any problems with this? If not we can merge it

raveit65 (approved):
LGTM

PR #303, opened 2023-05-05
Blacklist some session-specific variables (with corrected author name)

raveit65:
Same as https://github.com/mate-desktop/mate-session-manager/pull/282
But with corrected author name from gnome.

Things like XDG_SESSION_ID should not be uploaded to the environment.
For example this is broken currently:

  1. SSH to your machine
  2. Log in to MATE Shell
  3. Log out
  4. Log in again
  5. Lock the screen
  6. Try to unlock

You can't, and this is because the XDG_SESSION_ID from the first session
(step 2) has leaked through to the second one (step 4), and so MATE
Shell is listening to the `logind` `UnlockSession` signal for the wrong
session. The SSH session established in step 1 serves to keep the
`systemd --user` instance alive, so that the state is not torn down
between logins.

Patch from GNOME by Iain Lane <iainl@gnome.org>

### DIFF
--- a/mate-session/gsm-util.c
+++ b/mate-session/gsm-util.c
@@ -39,6 +39,13 @@
 
 static gchar *_saved_session_dir = NULL;
 
+static const char * const variable_blacklist[] = {
+    "XDG_SEAT",
+    "XDG_SESSION_ID",
+    "XDG_VTNR",
+    NULL
+};
+
 gchar **
 gsm_get_screen_locker_command (void)
 {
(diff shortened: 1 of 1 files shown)


zhuyaliang:
@raveit65 What should I do to ```Log in to MATE Shell``` and ```Log out```

raveit65:
I could never reproduced the original issue in bug report, (impossible to unlock the session after the session was locked).
But the problem is that after log out the session is still active.
```
[rave@mother ~]$ loginctl list-sessions
SESSION  UID USER SEAT  TTY  
      2 1000 rave seat0 
      5 1000 rave seat0 
      6 1000 rave       pts/7
      8 1000 rave seat0 

4 sessions listed.
[rave@mother ~]$ 
``` 
After every new login from login-manager (lightdm) a new session will create but the old one is still there.
But sadly this patch does fix nothing here with my fedora box.
Maybe it's different with debian or archLinux. 
Any way i will close this PR.

raveit65:
> @raveit65 What should I do to `Log in to MATE Shell` and `Log out`

I think login from Login-manager is meant.

Issue #304, opened 2023-05-20
Session hang during logout after upgrade to 1.26.1


jpalus:
#### Expected behaviour
Upon logout session closes cleanly.

#### Actual behaviour
Session remains open and memory error is logged in `.xsession-errors`:
```
malloc_consolidate(): unaligned fastbin chunk detected
```

#### Steps to reproduce the behaviour
1. Login
2. Logout

#### MATE general version
1.26

#### Package version
1.26.1

#### Linux Distribution
PLD

#### Link to bugreport of your Distribution (requirement)
None. Feel free to reject bugreport if that's really requirement.

PR #305, opened 2023-05-20
main: fix double free on gl_renderer

jpalus:
Fixes: 1263384 ("mate-session: fix memory leak")
Fixes: #304


jpalus:
Note that `gl_renderer` is freed at:
https://github.com/mate-desktop/mate-session-manager/blob/da643512befc61c0846bb399856e658630c5306d/mate-session/main.c#L778-L781

lukefromdc (approved):
No obvious problems on a new session with this applied but if there were it would be a memory leak not a crash. Checked the code, this was indeed a double free

PR #307, opened 2023-08-18
Add system desktop files item to automatic startup support

### LINKED ISSUE #150: [Feature Request] Startup Applications - Add .desktop items
Currently MATE Startup Applications requires the user to know the command path of the application. It would be more user friendly to add via an existing .desktop file so it shows the app name and icon.

Example from Budgie Settings:
image

zhuyaliang:
Fix #150


lukefromdc (approved):
Seems to work well. We have both an add and a custom add button with this. Custom add brings up the original dialog for starting arbitrary programs. The "add" dialog now gives a clickable `GtkList ` containing everything in /usr/share/applications, which should be everything in the application menus.  

This is the only place I know of where we are using the GtkList but it works well. I checked it in every theme I have using GtkInspector and it always looked good.

lukefromdc:
Hoping we can get one more review

capplet/msm-desktop-app-dialog.c
```diff
@@ -0,0 +1,311 @@
...
+
+    context = gtk_widget_get_style_context (searchbutton);
+    gtk_style_context_add_class (context, "image-button");
+    gtk_style_context_remove_class (context, "text-button");
+
+    headerbar = gtk_dialog_get_header_bar (GTK_DIALOG (dialog));
+    gtk_header_bar_pack_end (GTK_HEADER_BAR (headerbar), searchbutton);
+    g_object_bind_property (searchbutton, "active", searchbar, "search-mode-enabled", G_BINDING_BIDIRECTIONAL);
+}
+static GObject *
```
raveit65, line 162: A blank line before `static Gobjekt` would be nice.

zhuyaliang: done

raveit65:
This black area looks a bit weird, i think it is the GtkViewport inside the scrolled window.
mate-session-properties

zhuyaliang:
@raveit65 As expected, we need a border to look more aesthetically pleasing. Set here
```
G_ Object_ Set (dialog>listbox, "margin", 6, NULL);
```

raveit65:
I just comment out that line. For me it looks better
mate-session-properties2

So for what this margin is needed? 
Or better 1px margin?

raveit65:
I think the problem is the missing border-color from theme.
IHMO, it looks weird too with 1px margin.
mate-session-properties3

zhuyaliang:
removed margin

raveit65:
I guess what you mean is using `border-width` for the viewport, which is the parent of the listbox.
Changed with gtk-inspector.
Mate-session-properties4
Same with BlackMate
mate-properties5

zhuyaliang:
@raveit65 Fixed

raveit65:
Same as before. You added the border-width to GtkListBox. Here the border-width is still ignored.
mate-session-properties_7
What i meant is adding the border-width to the parent which is GtkViewport.
Screenshot with object hierarchy.
mate-session-properties_8

But for me this is only a little cosmetic change. I am fine with not using margin or border-width.

More important is that i am missing the search entry when not using headerbars.
General Function works fine and i could add firefox and thunderbird desktop files to autostart.
But i needed to scroll a long time to the end of the list to select thunderbird. 
Is the search-entry really only available for csd (headerbar) dialog?

And i am wondering of using `msm` term. Nearly all other functions or files are using `gsm` term.
I do not really care about and i don't want to block it when you think it is OK, but it looks like a bit inconsistent usage of naming.

raveit65:
Hmm, full functionality works only with:
```
[rave@mother ~]$ gsettings get org.mate.interface gtk-dialogs-use-header
true
```
mate-session-properties_9

But the gsettings key is disabled as default in MATE.

zhuyaliang:
@raveit65 Yes, the search function only works when using dialog headerbar mode，```gsm``` is the abbreviation for ```gnome session manager```, and ```msm``` is the abbreviation for ```mate session manager```. I think using ```msm``` is more appropriate

raveit65:
I can confirm that the border is there with latest forced-push.


> @raveit65 Yes, the search function only works when using dialog headerbar mode, .......

Mate default does not use headerbars for dialog windows. In 2014 we decided not use headerbars for Mate applications for several reasons.
https://github.com/mate-desktop/mate-desktop/pull/129
https://github.com/mate-desktop/mate-desktop/commit/e9ae72f43c0c5d859685c29a926284edb5455d4e
So most users of Mate will not see your searchbar.

Why not using a normal decorated dialog-window with a searchbar at the bottom?

When we want to use headerbars for dialog windows than this is a general design question and needs to be discussed by the team, ihmo.

Note:
Why is the searchbar not visible as default in headerbar-dialog? An extra click on the search button to show the searchbar is unneeded and one click too much for me ;-)

lukefromdc:
We should be able to pack that searchbox into a box at the top or bottom of the UI instead of the headerbar with little difficult. Similar to what I did for the dictionary applet in wayland, where we cannot not focus the entry if it is on the panel itself, so I moved it to the popup dialog.

Headerbars clash with the normal MATE server-side decorated UI which is why we do not use them by default. I am still looking for a way to disable them is wayland where for obvious reasons we don't have an xsettings manager

lukefromdc:
Doesn't seem to need any special theme support, looks quite nice in my own UbuntuStudio_Legacy theme

zhuyaliang:
searchbar  visible as default in headerbar-dialog

lukefromdc:
I can confirm we now get a visible searchbox underneath the non-CSD decorations

raveit65:
Thanks, both versions non-csd and csd dialog are working with with a active searchbar now.

capplet/msm-desktop-app-dialog.c
```diff
@@ -0,0 +1,314 @@
...
+    gtk_widget_set_size_request (GTK_WIDGET (dialog), 600, 400);
+    gtk_window_set_modal (GTK_WINDOW (dialog), TRUE);
+
+    dialog->entry = gtk_search_entry_new ();
+    g_signal_connect (dialog->entry,
+                     "search-changed",
+                      G_CALLBACK (on_search_entry_changed),
+                      dialog);
+
+    gtk_entry_set_placeholder_text (GTK_ENTRY (dialog->entry), _("Search Applications…"));
```
raveit65, line 118: The text isn't visible in non-csd version, and in headerbar vsersion you need to click on the button to make the text visible.

capplet/msm-desktop-app-dialog.c
```diff
@@ -0,0 +1,314 @@
...
+    sw = gtk_scrolled_window_new (NULL, NULL);
+    gtk_scrolled_window_set_policy (GTK_SCROLLED_WINDOW (sw),
+                                    GTK_POLICY_NEVER,
+                                    GTK_POLICY_AUTOMATIC);
+    gtk_container_add (GTK_CONTAINER (sw), dialog->listbox);
+    gtk_box_pack_start (GTK_BOX (content_area), sw, TRUE, TRUE, 0);
+
+    searchbutton = gtk_toggle_button_new ();
+    gtk_widget_set_valign (searchbutton, GTK_ALIGN_CENTER);
+    searchimage = gtk_image_new_from_icon_name ("edit-find-symbolic", GTK_ICON_SIZE_MENU);
```
raveit65, line 152: Any chance to use the colorful icon version? It's called `edit-find` ?

capplet/msm-desktop-app-dialog.c
raveit65: In non-csd dialog the text is visible when you drag the window o_0

capplet/msm-desktop-app-dialog.c
zhuyaliang: done

capplet/msm-desktop-app-dialog.c
raveit65: Got it, when the entry got focus than the text goes away.

raveit65 (approved):
LGTM

PR #316, opened 2025-03-18
Use g_info() for "screensaver left the bus" message

andrew-sayers:
This message is only useful for people developing screensavers, but occurs normally when the user logs out.

Change it to a g_info() message, so screensaver devs can find it but normal users don't think they have a bug.


lukefromdc (approved):
I don't know how to test this for function but it built without issue in a quick test build. I agree with moving development messages out of the warning category so as not to generate bug reports, spam logs, or confuse users.

yetist (approved):

PR #317, opened 2025-04-07
Fix unused variable warnings

yetist:
Fix unused variable warnings with `G_GNUC_UNUSED`.


lukefromdc:
This suppresses unused variable warnings while keeping the extra variables themselves. Is there a reason
such as maintaining ABI compatability with something else for this?

lukefromdc:
I would suggest the latter: less risk of typos sneaking into code, and normal builders who are not MATE developers
do not have any use for these warnings.

PR #320, opened 2025-06-02
Update window layout

luigifab:
Fix https://github.com/mate-desktop/mate-control-center/issues/785
Tested by rebuilding Mate 1.26 Debian source package.


lukefromdc:
We can ignore the travis issue. I got a clean build here, the math works (change 6px to 5px) and there is no way for this to breal anyone's build. Will merge

PR #326, opened 2026-01-01
Required components default fallback

### LINKED ISSUE #324: Reset window manager to default (if not found)  [feature request]
Suggestion: Reset the window manager to default, if not found.

Steps to reproduce:
- Changed with the window manager an external Tool (mate-tweak / mintdesktop).
- Uninstall this tool or the selected window-manager. => Window-manager is missing.

-------
The 2 situations in detail:
- Use of "mate-tweak" and changed window manager.
  After uninstall mate-tweak, the selected window manager is missing.
  Reason: Mate-tweak distribute it own marco-wrapper. ( Example: /usr/bin/marco-no-composite )
  An alternative solution would also fix it: Doing the reset with a post_uninstall script from mate-tweak ((I could write/submit one)).
- Use of "mintdesktop" and changed window manager (example: xfce4-wm).
  After uninstall xfce4-wm later, the selected window manager is missing.

-------
I've wrote a patch for "mate-wm". ( i realized afterward, that's the an unimportant place. ) :
```
# Reset to default, if executable of window manager not found
if [ -n "x$WINDOW_MANAGER" ] && [ "x$WINDOW_MANAGER" != "marco" ] && [ "x$WINDOW_MANAGER" != "xmate-wm" ] ; then
  if ! type "x$WINDOW_MANAGER" &> /dev/null ; then
    echo "Window manager $WINDOW_MANAGER executable not found. Will fallback and reset to default."
    gsettings reset org.mate.session.required-components windowmanager
    WINDOW_MANAGER="" # or "marco"
  fi
fi
```

2387skju:
Reset the `required components` to default, if not found. 
After uninstall a component, a fallback is need: will fix #324
_(( Required components are: dock, filemanager, panel, window manager. ))_

-------

Changes in file **mate-session/main.c** :
Reset will be done, in this case:
 - the executable was not found 
 - and the current value is not empty 
   - => this still allows the user to disable a component manual (= by set empty value) 
 - and the default value is not empty
 
-------
Additional just for the Windowmanager:
Changes in file **data/mate-wm** : 
- checks if the value from gsettings is usable (_`command`_) , 
- otherwise just ignore that value and continue the script normal (= will search for another window manager)

_(both shortly tested with older Mate Version 1.26 on Debian 13. ; I'm new in C-Language)_

### DIFF
--- a/data/mate-wm
+++ b/data/mate-wm
@@ -58,6 +58,14 @@ if [ "x$WINDOW_MANAGER" = "xcompiz" -o "x$DEFWM" = "xcompiz" ]; then
   fi
 fi
 
+#  Check WINDOW_MANAGER exists
+if [ -n "$WINDOW_MANAGER" ] ; then
+  if ! command -v "$WINDOW_MANAGER" > /dev/null 2>&1; then
+    echo "Window manager $WINDOW_MANAGER executable not found."
+    WINDOW_MANAGER=""
+  fi
+fi
+
 # Avoid looping if the session configuration tells us to use mate-wm or if
 # the user forces mate-wm via WINDOW_MANAGER
 if [ "x$WINDOW_MANAGER" = "xmate-wm" ]; then
--- a/mate-session/main.c
+++ b/mate-session/main.c
@@ -235,6 +235,56 @@ static void append_default_apps(GsmManager* manager, const char* default_session
 	g_strfreev (default_apps);
 }
 
+static void append_required_apps_add_component (GsmManager* manager, GSettings* settings_required_components, const char* component, gboolean is_recursive_call)
+{
+	char* default_provider;
+
+	default_provider = g_settings_get_string (settings_required_components, component);
+
+	g_debug ("main: %s looking for component: '%s'", component, default_provider);
+
+	if (default_provider != NULL)
+	{
+		char* app_path;
+
+		app_path = gsm_util_find_desktop_file_for_app_name(default_provider, NULL);
+
+		if (app_path != NULL)
+		{
+			gsm_manager_add_autostart_app(manager, app_path, component);
+		}
+		else
+		{
+			g_warning ("Unable to find provider '%s' of required component '%s'", default_provider, component);
+
+			if (default_provider[0] != '\0' && !is_recursive_call)
+			{
+				// possible reset component to default
+
+				const char *default_default_provider;
+				GVariant *g_settings_default_default;
+				g_settings_default_default = g_settings_get_default_value (settings_required_components, component);
+				default_default_provider = g_variant_get_string (g_settings_default_default, NULL);
+
+				if (default_default_provider[0] != '\0' && strcmp (default_default_provider, default_provider) != 0)
+				{
+					g_warning ("Reset required component '%s' to default", component);
+					g_settings_reset (settings_required_components, component);
+
+					append_required_apps_add_component(manager, settings_required_components, component, TRUE);
+				}
+
+				g_variant_unref (g_settings_default_default);
+			}
+
+		}
+
+		g_free(app_path);
+	}
+
+	g_free(default_provider);
+}
+
 static void append_required_apps(GsmManager* manager)
 {
 	gchar** required_components;
@@ -257,39 +307,11 @@ static void append_required_apps(GsmManager* manager)
 	{
 		for (i = 0; required_components[i]; i++)
 		{
-			char* default_provider;
-			const char* component;
-
 			if (IS_STRING_EMPTY((char*) required_components[i]))
 			{
 				continue;
 			}
-
-			component = required_components[i];
-
-			default_provider = g_settings_get_string (settings_required_components, component);
-
-			g_debug ("main: %s looking for component: '%s'", component, default_provider);
-
-			if (default_provider != NULL)
-			{
-				char* app_path;
-
-				app_path = gsm_util_find_desktop_file_for_app_name(default_provider, NULL);
-
-				if (app_path != NULL)
-				{
-					gsm_manager_add_autostart_app(manager, app_path, component);
-				}
-				else
-				{
-					g_warning("Unable to find provider '%s' of required component '%s'", default_provider, component);
-				}
-
-				g_free(app_path);
-			}
-
-			g_free(default_provider);
+			append_required_apps_add_component(manager, settings_required_components, required_components[i], FALSE);
 		}
 	}
 
--- a/mate-session/main.c
+++ b/mate-session/main.c
@@ -235,6 +235,60 @@ static void append_default_apps(GsmManager* manager, const char* default_session
 	g_strfreev (default_apps);
 }
 
+static void append_required_apps_add_component (GsmManager* manager, GSettings* settings_required_components, const char* component, gboolean is_recursive_call)
+{
+	char* default_provider;
+
+	default_provider = g_settings_get_string (settings_required_components, component);
+
+	g_debug ("main: %s looking for component: '%s'", component, default_provider);
+
+	if (default_provider != NULL)
+	{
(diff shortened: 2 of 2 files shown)


vkareh (changes requested):

data/mate-wm
```diff
@@ -43,6 +43,13 @@ done
 # WINDOW_MANAGER overrides all
 if [ -z "$WINDOW_MANAGER" ] ; then
     WINDOW_MANAGER=`gsettings get org.mate.session.required-components windowmanager 2> /dev/null | sed -r -e "s/^'([^']*)'$/\\1/"`
+    if [ -n "$WINDOW_MANAGER" ] ; then
+        if ! command -v "$WINDOW_MANAGER" > /dev/null 2>&1; then
+           # Known bug: xcompiz / "Migrate compiz to compiz-manager" (=next block) is now not working anymore.
```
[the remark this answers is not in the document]
2387skju: Fixed: I've put the code to a bedder position.
(( = No regression anymore. It was only introduced by the wrong position of this new code. ))

mate-session/main.c
```diff
@@ -283,7 +283,36 @@ static void append_required_apps(GsmManager* manager)
 				}
 				else
 				{
-					g_warning("Unable to find provider '%s' of required component '%s'", default_provider, component);
+					// possible reset component to default
+					const gchar *gsetting_default_default_provider;
+					const GVariant *default_default_provider;
+					default_default_provider = g_settings_get_default_value (settings_required_components, component);
+					g_variant_get (default_default_provider, "&s", &gsetting_default_default_provider);
+					g_free (default_default_provider);
```
vkareh: should be `g_variant_unref`

mate-session/main.c
```diff
@@ -283,7 +283,36 @@ static void append_required_apps(GsmManager* manager)
 				}
 				else
 				{
-					g_warning("Unable to find provider '%s' of required component '%s'", default_provider, component);
+					// possible reset component to default
+					const gchar *gsetting_default_default_provider;
+					const GVariant *default_default_provider;
+					default_default_provider = g_settings_get_default_value (settings_required_components, component);
+					g_variant_get (default_default_provider, "&s", &gsetting_default_default_provider);
```
vkareh: This `&s` returns a pointer into the `default_default_provider` internal buffer, so when you free the gvariant in the next line, every use of `gsetting_default_default_provider` (all those `strcmp` below) is reading memory that has been freed already...

I think the solution is to either use `s` and then free `g_free(gsetting_default_default_provider)` when you're done with it, or alternatively _not_ free `default_default_provider` until you're done with `gsetting_default_default_provider`.

vkareh (changes requested):
There are some memory management issues, but this is moving in the right direction, thanks.

2387skju:
Thanks for helping :-)

### Changes:
 - variable name swapped each other:
   - `g_settings_default_default` <=> `default_default_provider`
   - This is more consistent to the origin code (the lines before).
 - replaced:
   - old: `strcmp(myString, "") != 0`
   - new: `myString[0] != '\0'`
    - The return of [`g_settings_get_string`](https://docs.gtk.org/gio/method.Settings.get_string.html) and [`g_variant_get_string`](https://docs.gtk.org/glib/method.Variant.get_string.html) is NUL terminated. ( NULL is not mentioned here in glib doku.)
 - replaced:
    - old: [`g_variant_get`](https://docs.gtk.org/glib/method.Variant.get.html)
    - new: [`g_variant_get_string`](https://docs.gtk.org/glib/method.Variant.get_string.html)
    - This is compare-able like `&s` version of `g_variant_get`.
    - I think this look a little bit bedder. (but not important)

-----

### Split code in a new function ? :
A suggestion:
Just an Idea. I'm not sure about it. The benefit is not very deep, in total just a few lines less.
In this structure:
```
append_required_apps_add_component (...,  gboolean is_recursive_call)
{

  [...] // code do add component
  
  if  // if failed && ! is_recursive_call
  {
    // code do reset
    ...
    append_required_apps_add_component (..., TRUE);
  }
}

append_required_apps ()
{
  for  // for each required_components
  {
    append_required_apps_add_component (..., FALSE);
  }
}
```

This function implemented in Code:
 - This [branch](https://github.com/2387skju/mate-session-manager/tree/required_components_default_fallback__in_function) or [commit](https://github.com/2387skju/mate-session-manager/commit/b4ad92832f03703cd2fb7e1de77701d4a3902146)
 - ((( or the same, with an extra refactor commit: [branch](https://github.com/2387skju/mate-session-manager/tree/required_components_default_fallback__in_function__with_refactor_commit) or [commit](https://github.com/2387skju/mate-session-manager/commit/1055ae4ec502115e1f440bfd6209a20b9c4290fa) ))

mate-session/main.c
2387skju: I've tested both solutions, and both are working.
But: I've found another suitably function, which a prefer (( just a little bit more )):
  - ((( note: variable name was changed / swapped )))
  - changed from [`g_variant_get`](https://docs.gtk.org/glib/method.Variant.get.html) => [`g_variant_get_string`](https://docs.gtk.org/glib/method.Variant.get_string.html)

vkareh:
@2387skju this looks a lot better, thanks! I think you have a good suggestion there with the function, it will make things easier to understand for sure.

mate-session/main.c
```diff
@@ -235,6 +235,56 @@ static void append_default_apps(GsmManager* manager, const char* default_session
...
+		{
+			g_warning ("Unable to find provider '%s' of required component '%s'", default_provider, component);
+
+			if (default_provider[0] != '\0' && !is_recursive_call)
+			{
+				// possible reset component to default
+
+				const char *default_default_provider;
+				GVariant *g_settings_default_default;
+				g_settings_default_default = g_settings_get_default_value (settings_required_components, component);
```
vkareh: Should probably add a null check here

mate-session/main.c
```diff
@@ -235,6 +235,56 @@ static void append_default_apps(GsmManager* manager, const char* default_session
...
+		else
+		{
+			g_warning ("Unable to find provider '%s' of required component '%s'", default_provider, component);
+
+			if (default_provider[0] != '\0' && !is_recursive_call)
+			{
+				// possible reset component to default
+
+				const char *default_default_provider;
+				GVariant *g_settings_default_default;
```
vkareh: Can we rename this variable? It reads too much like a GLib API function. Maybe `default_variant` or `schema_default` would work better.

vkareh (commented):
This is looking great! Just some minor nitpicks before this is ready

2387skju:
Sorry, I messed up a commit and tried to clean it. Please just take a look at the last Diff-Compare in GitHub: This one is from before last review f9f03f8 to the current state 98da258.

L-U-T-i:
I am getting a build error with the latest patch:
```
main.c: In function 'append_required_apps_add_component':
main.c:283:50: error: 'g_settings_default_default' undeclared (first use in this function); did you mean 'gtk_settings_get_default'?
  283 |                                 g_variant_unref (g_settings_default_default);
      |                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                  gtk_settings_get_default

```
It seems to me you've renamed `g_settings_default_default` to `schema_default`, but you've forgotten to replace it here.

After I've changed:
`g_variant_unref (g_settings_default_default);`
to:
`g_variant_unref (schema_default);`
the build went through as expected.

2387skju:
@L-U-T-i: Thank you for helping :-)

I've just forget to delete this line in my committed version.

_Here you see a bedder human readable, corrected [git history of the last changes](https://github.com/2387skju/mate-session-manager/commits/required_components_default_fallback__tmp_change_reconstruction/)_

PR #330, opened 2026-02-04
session: Add sound events for login and logout

### LINKED ISSUE #328: Add the ability for a custom shutdown sound
I know it might be a redundant or rather unnecessary feature, but having it along with a startup sound through "paplay" would be a really amazing addition for me at least. It'd play right after choosing to reboot or shutdown, and before MATE kills the shell and starts the process. Same would go for logging out too, with suspending excluded from this rule. It'd be appreciated to have this feature since system sounds in general seem like a hit or miss too.

vkareh:
Adds optional libcanberra support to play desktop-login and desktop-logout sound events during session start and end. Sound playback is controlled through the system sound theme configuration.

Fixes #328

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -65,6 +65,10 @@
 #endif
 #include "gsm-session-save.h"
 
+#ifdef HAVE_LIBCANBERRA
+#include <canberra-gtk.h>
+#endif
+
 #define GSM_MANAGER_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), GSM_TYPE_MANAGER, GsmManagerPrivate))
 
 #define GSM_MANAGER_DBUS_PATH "/org/gnome/SessionManager"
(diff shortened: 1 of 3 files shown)


lukefromdc (approved):
While system sounds do not work on my setup, I do have libcanberra installed and a build of this finished with no problems.
The config log showed
`Libcanberra support:      yes`
So that part I know works. Hoping we can get another review but if it works for you I am OK with merging. The code change is simple enough.

AngryDavid808:
I thank you so much for the implemetation. I understand it was cosmetic mostly, but it was a change that's well-deserved for such an amazing DE. Keep up the great work ✨

mate-session/gsm-manager.c
```diff
@@ -1548,6 +1552,13 @@ do_phase_query_end_session (GsmManager *manager)
         data.flags = 0;
         priv = gsm_manager_get_instance_private (manager);
 
+#ifdef HAVE_LIBCANBERRA
+        ca_context_play (ca_gtk_context_get (), 0,
+                         CA_PROP_EVENT_ID, "desktop-logout",
+                         CA_PROP_EVENT_DESCRIPTION, "Session logout",
+                         NULL);
+#endif
```
cwendling, line 1560: Is that really the right place for this? I admittedly didn't test this, but it looks like it should rather be in `GSM_MANAGER_PHASE_END_SESSION` rather than `GSM_MANAGER_PHASE_QUERY_END_SESSION`, shouldn't it?  Or is it too late then?
The thing I'm mostly "afraid" of is what happens if some apps inhibit session close?

vkareh: Oh crap, you're right. Like if an unsaved document prevents the session from ending the sound would end up playing even if the user decides to cancel the logout at that point. I'll move that.

vkareh: Here you go: https://github.com/mate-desktop/mate-session-manager/pull/333

PR #333, opened 2026-03-19
session: Move logout sound to end-session phase

vkareh:
The desktop-logout sound event was being played during the query-end-session phase, which means it would play even if an application inhibited the logout and the user cancelled. Move it to the end-session phase where logout is already confirmed.

### DIFF
--- a/mate-session/gsm-manager.c
+++ b/mate-session/gsm-manager.c
@@ -1552,13 +1552,6 @@ do_phase_query_end_session (GsmManager *manager)
         data.flags = 0;
         priv = gsm_manager_get_instance_private (manager);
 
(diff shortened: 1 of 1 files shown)


AngryDavid808 (commented):
I'll have to compile and test again later, will post back results once I do so. But yes, the update before that was not playing a logout sound or a login sound either, thought I'd post back those results. I used Linux Mint (latest) for the test last time.

mbkma (approved):
This change makes sense. Moving the logout sound to the end-session phase avoids the false-positive case where the sound plays even though logout is later cancelled, and it keeps the behavior aligned with the point where the session is actually going away.

PR #334, opened 2026-04-01
session: restart non-client apps that set AutoRestart

### LINKED ISSUE #321: X-MATE-AutoRestart doesn't work for non-client apps
Hi! I'm the new maintainer of gnome-session upstream, and I found a bug that mate-session-manager seems to have inherited since it forked. So I'm forwarding the bug report here

Some apps like the Orca screen reader set X-GNOME-AutoRestart with the expectation that they'll get restarted in the event of a crash or unclean exit. However, the gnome-session codebase you're forked from doesn't implement this for plain apps, only clients. A client is an app that talks to the session manager via dbus or via XSMP. Orca does neither, and so X-GNOME-AutoRestart simply doesn't work. I think this wasn't an intended design because someone clearly thought of supporting it in gsm-autostart-app.c: see the "died" signal that gets emitted. It's just that nothing subscribes to that signal.

See https://gitlab.gnome.org/GNOME/orca/-/issues/622 and https://gitlab.gnome.org/GNOME/orca/-/merge_requests/255 for some more details.

In your case, it also appears that you renamed the X-GNOME-AutoRestart setting to X-MATE-AutoRestart. This further breaks a case like Orca's, which sets X-GNOME-AutoRestart with the expectation that gnome-session-compatible session managers (like mate-session-manager, cinnamon-session, and unity-session) will continue to pick them up. unity-session (which is just gnome-session 48) and [cinnamon-session](https://github.com/linuxmint/cinnamon-session/blob/master/cinnamon-session/csm-autostart-app.h#L69) do listen to X-GNOME-AutoRestart (and other X-GNOME-* settings). I'd suggest checking for X-MATE-* settings, and if they're not present falling back to X-GNOME-* settings for backwards-compatibility.

vkareh:
Apps like Orca set `X-GNOME-AutoRestart=true` in their .desktop files expecting to be restarted after a crash. However, the restart logic only handled apps that registered as XSMP or D-Bus clients. Non-client apps emitted "died"/"exited" signals but nothing subscribed to them. This bug was inherited from gnome-session.

Connect the "died" and "exited" signals on autostart apps to trigger restart for non-client apps with AutoRestart enabled. Also fall back to `X-GNOME-AutoRestart` when `X-MATE-AutoRestart` is not present, for compatibility with apps that only set the GNOME key.

Fixes #321

### DIFF
--- a/mate-session/gsm-autostart-app.c
+++ b/mate-session/gsm-autostart-app.c
@@ -474,7 +474,18 @@ load_desktop_file (GsmAutostartApp *app)
                                                                   GSM_AUTOSTART_APP_AUTORESTART_KEY,
                                                                   NULL);
         } else {
-                priv->autorestart = FALSE;
+                /* Fall back to X-GNOME-AutoRestart for compatibility with
+                 * apps like Orca that set the GNOME key */
+                res = egg_desktop_file_has_key (priv->desktop_file,
+                                                "X-GNOME-AutoRestart",
+                                                NULL);
+                if (res) {
+                        priv->autorestart = egg_desktop_file_get_boolean (priv->desktop_file,
+                                                                          "X-GNOME-AutoRestart",
+                                                                          NULL);
+                } else {
+                        priv->autorestart = FALSE;
+                }
         }
 
         g_free (priv->condition_string);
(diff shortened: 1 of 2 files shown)


mbkma (approved):
Good fix. Hooking the restart path up for non-client apps closes the gap here, and the extra guard against restarting the same app twice when it also has a registered client is exactly the right thing to keep.

PR #337, opened 2026-06-01
app: Add respawn rate limiting to prevent CPU spinning

### LINKED ISSUE #336: high cpu usage (recent regression)
### Steps to Reproduce

run top

### Expected Behavior

about zero cpu usage

### Actual Behavior

seems one core busy

### Error Output

```shell

```

### Version

6c53bde758e6d8944438c2469b8a58209db6d0b2

### Distribution

debian trixie aarch64, this issue are reproducible with debian forky x64 (but DE looks unusable there for other reasons)

### Additional Context

first bad commit: [8e1cfe174592f1abe329956f7946f4dd29bd6652] session: restart non-client apps that set AutoRestart

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

vkareh:
When a non-client app with AutoRestart enabled exits immediately after starting, the restart logic would loop without delay, causing high CPU usage. This adds a 60-second rate limit to gsm_app_restart. If an app respawns twice within this window, the second restart is rejected.

The code is adapted from:
- https://gitlab.gnome.org/GNOME/gnome-session/-/commit/4df48234
- https://gitlab.gnome.org/GNOME/gnome-session/-/commit/b9aa675e

Fixes #336

### DIFF
--- a/mate-session/gsm-app.c
+++ b/mate-session/gsm-app.c
@@ -30,11 +30,15 @@
 #include "gsm-app.h"
 #include "org.gnome.SessionManager.App.h"
 
+/* If a component crashes twice within a minute, we count that as a fatal error */
+#define _GSM_APP_RESPAWN_RATELIMIT_SECONDS 60
+
 typedef struct {
         char            *id;
         char            *app_id;
         int              phase;
         char            *startup_id;
+        gint64           last_restart_time;
         GDBusConnection *connection;
         GsmExportedApp  *skeleton;
 } GsmAppPrivate;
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
I cannot test run this at the moment as I am in the middle of a mate-wayland session, but the code itself looks good. If the crashing app has an existing restart time (has restarted before) within the rate limit time, `gsm_app_restart` returns FALSE 

A build of this finished fine, didn't even see any warnings from mate-session-gsm-app.o

I don't have any non-MATE apps that use this that I know use it so not really up for a good functional test but this has sat for a while and the CPU sitting and spinning is something we need to fix

lukefromdc:
I just tested this in an X11 session, with a non-working orca install set to start witb the session. Didn't get an excess CPU spin with this, and the code looks good to me

